PR #1, opened 2024-03-18
mrc-5134 Basic app

EmmaLRussell:
Proof of concept for the DengueMap app, which will potentially be generalised to other arboviruses, hence the repo name. 

This branch currently deployed on GH pages here: https://mrc-ide.github.io/arbomap/

The basic functionality of the full app is that the user will initially see choropleth for all available countries at admin level 1. They can click/tap to select which indicator to show values for. They can click on a country to select it - this will zoom to that country, and display data for admin level 2. They can click again to unselect the country and zoom back out.  Hovering (press and hold on touchscreen) over a region will show tooltip with details. 

The proof of concept has geojson and dummy data for two countries (Malawi and Tanzania). You can click to select either country - there's then a bit of an alarming pause while admin2 data is loaded (loading spinner tbd!), and you can click again to unselect. Autozoom is not implemented yet. There are tooltips, currently unformatted, and dummy values for 'FOI' and 'p9', which you can switch between - friendly indicator names are also tbd. 

This is a pure front end app using Vuetify (https://vuetifyjs.com), and scaffolded with `create-vuetify`. The Choropleth uses d3 colour scales and is largely nicked from HINT. 

Geojson (sourced from [geoboundaries](https://www.geoboundaries.org/simplifiedDownloads.html)) and indicator data files are loaded from `/public/resources` as is a `config.json` file listing supported countries and other config. 

There's lots of scaffolding cruft and demo data here, most of the real code is in:
- `src/stores/appStore.ts`: the Pinia store for the app
- `src/components/Choropleth.vue`: the map component
- `src/composables/useColourScale.ts`: colour scale functionality for the Choropleth
- `src/pages/index.vue`: index page, containing map component and the indicator buttons
- `src/resources/utils.ts`: utils for loading resource files

There is linting, but no tests yet. 

Still TODO (to be made into tickets):
- add tests for proof of concept functionality: mrc-5191
- Support partial country data - we expect NAs for some Australian regions: mrc-5199
- Use real dataset and geojson for all countries: mrc-5144, mrc-5167, mrc-5166
- Rename this repo to arbomap-dev and make separate arbomap repo for production deployment. Make a workflow to do deployment from dev to prod on release. Note that this will require some config changes. mrc-5198
- Obtain external domain and associate with Github pages deployment: mrc-5197
- Map bounds to lock to selected country boundary on select. (Not sure what should happen on unselect? Snap out to global view, or last user selected bounds?) mrc-5169
- Load admin1 regions progressively, from central longitude: mrc-5201
- Make sure data is cached in browser: mrc-5202
- Display country outlines, and admin1 outlines when view admin2 data: mrc-5203
- (Maybe) generate more simplified admin1 boundaries: mrc-5204
- Loading spinner when loading data: mrc-5205
- Sort out map height - currently falling off the bottom of the viewport: mrc-5206
- Excel download - selected country only, with admin1 and admin2 values on separate tabs: 5170
- Use friendly indicator names - get these from Ilaria and Clare - and include indicator explanations somewhere. Will probably need to overhaul indicator selection at this point. Maybe use expandable view, maybe with nice icons showing colour scales? mrc-5207
- Format tooltips - round and present numbers in sensible way (2dp or 2sf), show selected indicator first and styled more prominently: mrc-5208
- Routing - Put the current app under /arbomap/dengue: mrc-5196
- Routing - make routes for country and indicator e.g. /dengue/foi/mwi to load foi indicator, with MWI as selected country. Could omit country to select foi indicator with no selected country. Can navigate direct to these routes, and using the app controls should also navigate as routes, should be able to use back button through previous selections: mrc-5196
- Support multiple versions of data - optional route part e.g. /dengue/2024/foi/mwi (hmm, this potentially makes clashing routes if both country and version are optional...). Omitting version will always show latest data: mrc-5209
- Attribution for GADM: mrc-5210
- Map image download: mrc-5171
- About page, with logos (get text from Ilaria and Clare. Ilaria has sent some of the logos.) mrc-5173
- Privacy page: mrc-5211
- Accessibility page: mrc-5212
- Analytics - use an alternative to Google Analytics mrc-5213
- Instructions to user - tap/click to select and unselect, press and hold/hover to see region details mrc-5214
- Additional colour scale options - it's currently just scaling to all visible regions. We should configure min and max values for all indicators, and either choose sensible defaults (0-1) or use values generated from the real data i.e. actual min and max values across all regions at both levels: mrc-5172
- Selectable theme - just dark at the moment: mrc-5215
- Set global bounding box for unselected view: mrc-5216

...and probably some other stuff I haven't thought of!


src/auto-imports.d.ts
```diff
@@ -0,0 +1,191 @@
+/* eslint-disable */
```
richfitz, line 1: is this generated from somewhere? You can add it to .gitattributes to mark it as such for nicer diffs https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

```
src/*.d.ts linguist-generated=true
```

but it does make me wonder how it is generated and if I'd need to do it again?

src/composables/useColourScale.ts
```diff
@@ -0,0 +1,81 @@
...
+import { useAppStore } from "../stores/appStore";
+
+interface IndicatorRange {
+    min: number;
+    max: number;
+    range: number;
+}
+
+export const useColourScale = (selectedIndicators: Ref<Dict<FeatureIndicatorValues>>) => {
+    // TODO: we currently just scale colours to min and max in data, but
```
richfitz, line 15: I suspect this will be the desired behaviour, possibly right off the bat. Prevelance will naturally scale from 0 to 1 (though they may want more detail at the lower end). FOI is harder because  it's effectively a rate and so has no upper bound. We'd need to make sure that we're scaling against the largest value globally (so across all ADM2 units in all data sets) and that will be easiest to get as part of some sort of processing.

Can be changed later though!

src/layouts/README.md
```diff
@@ -0,0 +1,5 @@
+# Layouts
+
+Layouts are reusable components that wrap around pages. They are used to provide a consistent look and feel across multiple pages.
```
richfitz, line 3: The word "wrap" here made me wonder what happens if the map wraps; turns out that if you go all the way around the world the coloured bit goes away! This sounds innocent enough but might create some visual weirdnesses and definitely calls for an eventual reset view button.  Better would be for the geojson to display at longiude 540 etc; there might be a leaflet option for this too. (another for later!)

src/pages/index.vue
```diff
@@ -0,0 +1,34 @@
+<template>
+    <Choropleth />
+    <div class="sticky-footer">
+        <div v-for="name in indicatorNames" :key="name">
+            <v-btn
+                link
+                class="mb-2"
+                :class="name === selectedIndicator ? 'bg-blue' : 'bg-black'"
```
richfitz, line 8: also grey text? otherwise, especially with two indicators it's not totally obvious which is active

src/components/Choropleth.vue
```diff
@@ -0,0 +1,152 @@
...
+
+// TODO: pull out tooltips stuff into composable when fully implement
+const tooltipForFeature = (feature: Feature) => {
+    let indicatorValues = "";
+    const fid = featureId(feature);
+    if (fid in selectedIndicators.value) {
+        const featureValues = selectedIndicators.value[fid];
+        indicatorValues = Object.keys(featureValues)
+            .map((key) => {
+                return `${key}: ${featureValues[key].mean} (+/- ${featureValues[key].sd})<br/>`;
```
richfitz, line 108: Round the numbers here to 2dp or 2sf perhaps?

Possibly (later?) indicate which is the indicator visible by styling it differently?

package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "dengue-map",
+  "version": "0.0.0",
+  "scripts": {
+    "dev": "vite",
+    "build": "vue-tsc --noEmit && vite build",
+    "preview": "vite preview",
+    "lint-fix": "eslint . --fix --ignore-path .gitignore",
+    "lint": "eslint . --ignore-path .gitignore"
```
absternator, line 12: would add prettier script too for formats and lint.. these can be added to CI 
```
    "format:check": "prettier --check src/",
    "format:fix": "prettier --write src/",
    ```

File: package.json [hunk 7]
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "dengue-map",
+  "version": "0.0.0",
+  "scripts": {
```
absternator, line 7: lock node version so people run install with the correct version 

```suggestion
  
``` "engines": {
    "node": "^20.0.0"
  },

File: src/stores/appStore.ts [hunk 8]
```diff
@@ -0,0 +1,110 @@
+import { defineStore } from "pinia";
+import { Feature } from "geojson";
+import { getAppConfig, getGeojson, getIndicators } from "../resources/utils";
+import { AppConfig } from "../resourceTypes";
+import { FeatureGeojson, FeatureIndicators } from "../types";
+
+export const useAppStore = defineStore("app", {
+    state: () => {
```
absternator: can use implicit return here... avoids having to declare vars and returning at end... `state: () => ({loading: true, selectedIndicator: "" })`

src/stores/appStore.ts
```diff
@@ -0,0 +1,110 @@
...
+        const selectedIndicator = "";
+
+        // When a country is selected, we display admin2 data for it.
+        // We display admin1 data for all other countries
+        const selectedCountryId = "";
+
+        const appConfig: AppConfig | null = null;
+
+        // We keep all data in dictionaries with country ids as keys, mirroring the resources on disk
+        const admin1Indicators: Dict<FeatureIndicators> = {};
```
absternator: what is admin1 and admin2...

src/stores/appStore.ts
```diff
@@ -0,0 +1,110 @@
...
+            admin2Indicators,
+            admin2Geojson
+        };
+    },
+    getters: {
+        selectedIndicators: (state): FeatureIndicators => {
+            const { selectedCountryId, admin1Indicators, admin2Indicators } = state;
+            // get single dictionary of feature id to indicator values for user selections
+            if (!selectedCountryId) {
+                return Object.assign({}, ...Object.values(admin1Indicators));
```
absternator, line 29:
```suggestion
                return {...Object.values(admin1Indicators)}
```

src/stores/appStore.ts
```diff
@@ -0,0 +1,110 @@
+import { defineStore } from "pinia";
+import { Feature } from "geojson";
+import { getAppConfig, getGeojson, getIndicators } from "../resources/utils";
+import { AppConfig } from "../resourceTypes";
+import { FeatureGeojson, FeatureIndicators } from "../types";
```
absternator: `FeatureGeojson` is not in types file?

src/stores/appStore.ts
```diff
@@ -0,0 +1,110 @@
...
+            // admin2 indicators for the selected country
+            // NB we could return all indicators and look up values from features for the purposes of the map,
+            // but it will be useful to directly access all selected indicators when exporting data.
+            const filteredAdmin1 = Object.entries(admin1Indicators)
+                .filter(([countryId]) => countryId !== selectedCountryId)
+                .map(([, indicators]) => indicators);
+
+            return Object.assign({}, admin2Indicators[selectedCountryId], ...filteredAdmin1);
+        },
+        selectedFeatures: (state): Feature[] => {
```
absternator, line 42: whats a *feature*? sorry my lack of knowlede here

src/stores/appStore.ts
```diff
@@ -0,0 +1,110 @@
+import { defineStore } from "pinia";
+import { Feature } from "geojson";
+import { getAppConfig, getGeojson, getIndicators } from "../resources/utils";
+import { AppConfig } from "../resourceTypes";
```
absternator: *resoureTypes* file dosent exist?

src/stores/appStore.ts
```diff
@@ -0,0 +1,110 @@
...
+import { FeatureGeojson, FeatureIndicators } from "../types";
+
+export const useAppStore = defineStore("app", {
+    state: () => {
+        const loading = true;
```
absternator: initial loading state should probably be false

src/types.ts
absternator, line 1: should probably create *types* folder and split this up a bit into different files.. may lead to 1 huge file in the future

src/components/Choropleth.vue
```diff
@@ -0,0 +1,152 @@
...
+
+const map = ref<typeof LMap | null>(null);
+const featureRefs = ref<(typeof LGeoJson)[]>([]);
+
+const { selectedFeatures, selectedIndicators, loading, selectedIndicator } = storeToRefs(useAppStore());
+const { selectCountry } = useAppStore();
+
+const { colourScales, getColour } = useColourScale(selectedIndicators);
+
+const featureId = (feature: Feature) => feature.properties![FEATURE_ID_PROP];
```
absternator: funtions should probably be action words.. eg. `getFeatureId`

src/components/Choropleth.vue
```diff
@@ -0,0 +1,152 @@
...
+        return {
+            feature,
+            colour: getColourForFeature(feature, selectedInd)
+        };
+    });
+});
+
+const updateBounds = () => {
+    if (!loading.value) {
+        if (map.value && map.value.leafletObject) {
```
absternator:
```suggestion
        if (map.value?.leafletObject) {
```

absternator (commented):
Looks great awsome work !! Just had some small comments and have some questions for me and some suggestions 
1. what is admin1 and admin2 values? 
2. what is FOI / P9? maybe some tooltips would be good on the legend select
3. A small help bubble would be great how to use
4. also probs best to use `Record<string, T>` over the `Dict` type for dictionaries... i think maybe depricated typescript type because i cant even find in docs

src/auto-imports.d.ts
EmmaLRussell: Yeah, they're auto-generated on build. Yes. will add to .gitattributes, that will definitely make the PRs cleaner.

src/layouts/README.md
EmmaLRussell: Ah right, I did set a min zoom level to try to avoid some of this, but yes, you could pan all the way round and lose the regions when you get back to where you were. Yeah, should be able to set outer bounding box I think, in the same way we should be able to when zoomed to a country.

src/pages/index.vue
EmmaLRussell: Yeah something like that - had some chat with @M-Kusumgar about nice ways to show this - it might be nice to generate a little svg icon like in Google maps showing the colour scale for each indicator (some arrangement of blobs with selected colours from the scale). Maybe put a border round the selected one or something. We may also have to deal with expanding the indicators view once they get more numerous/verbose.

EmmaLRussell:
>1. what is admin1 and admin2 values?

Yeah, sorry should have explained! Admin 1 and admin 2 are regional levels for which data is collected and estimates are made. Admin1 is the biggest sub-national level, and admin2 is the next level down - so maybe corresponding roughly to 'region' and 'county'. We're only working with admin1 and admin2 levels for this project. 

> 
>2. what is FOI / P9? maybe some tooltips would be good on the legend select

Force of Infection, and prevalence in under 9s. Yup, will definitely be adding more information and friendly names for the indicators in a future ticket. 

>3. A small help bubble would be great how to use

Already on the TODO list! :)

>4. also probs best to use `Record<string, T>` over the `Dict` type for dictionaries... i think maybe depricated typescript type because i cant even find in docs

Yeah, I noticed intellisense was complaining! I might just define Dict in a types module as `Record<string, T` as it's nice and terse...

package.json
EmmaLRussell: I've already included prettier in the linting rules, rather than making it a separate process. I think it should amount to the same thing shouldn't it?

src/stores/appStore.ts
EmmaLRussell: Yeah, will do. I'll add an `AppState` interface too, so it's clear what type everything needs to be.

src/stores/appStore.ts
EmmaLRussell: Unfortunately this doesn't quite get us what we want, because Object.values is an array - what you actually end up with is an object with keys corresponding to the array indexes (so effectively replacing the country keys with those indexes as keys) rather than flattening the whole dictionary down to feature level.

src/stores/appStore.ts
EmmaLRussell: Lol, yeah, had some dead types hanging around, and not being highlighted by the linter. Should be fixed now!

src/stores/appStore.ts
EmmaLRussell: A feature is a geojson element, defined in the geojson files. It's basically an array of 2D points on the global map which defines the shape of the region, and where it is in the world, along with some properties which provide metadata like name and id.

src/stores/appStore.ts
EmmaLRussell: updated!

src/stores/appStore.ts
EmmaLRussell: I think it's OK for that to initialise to true because we certainly will be loading the admin1 data when the app first loads - though we may tweak that later if we add incremental loading of admin1 data.

src/types.ts
EmmaLRussell: Good idea!

src/components/Choropleth.vue
EmmaLRussell: Agreed

richfitz:
I've just moved this repo to `mrc-ide/arbomap-dev`, which at some point is going to break the github pages url

richfitz:
Actually, just backed that out because we will need the root url config option in before we have multiple deployments:

* `arbomap-dev/` (this repo eventually)
* `arbomap/` (the prod repo before we get the URL purchased)
* `/` (once we have the URL purchased)

src/components/Choropleth.vue
```diff
@@ -0,0 +1,152 @@
+<template>
+    <div v-if="initialising">loading..</div>
+    <div v-else>
+        <LMap ref="map" style="height: 100vh; width: 100%" @ready="updateBounds">
+            <LTileLayer
+                :url="backgroundLayer.url"
+                :attribution="backgroundLayer.attribution"
+                :max-zoom="backgroundLayer.maxZoom"
+                :min-zoom="backgroundLayer.minZoom"
+            ></LTileLayer>
```
M-Kusumgar: i think
```suggestion
            <LTileLayer v-bind="backgroundLayer"></LTileLayer>
```

src/components/Choropleth.vue
```diff
@@ -0,0 +1,152 @@
...
+
+const getFeatureId = (feature: Feature) => feature.properties![FEATURE_ID_PROP];
+const getFeatureName = (feature: Feature) => feature.properties![FEATURE_NAME_PROP];
+
+const getColourForFeature = (feature, indicator) => {
+    const featureIndicators = selectedIndicators.value[getFeatureId(feature)];
+    return getColour(indicator, featureIndicators);
+};
+
+// TODO: This structure is an inefficient way to get reactivity working, but we can probably do better!
```
M-Kusumgar: what do you mean by this?

M-Kusumgar (approved):
Happy to approve this, had a tiny suggestion and question!

absternator (approved):
Nice work LGTM

richfitz (approved):

src/components/Choropleth.vue
EmmaLRussell: Great thanks, I forgot you could do this! :)

src/components/Choropleth.vue
EmmaLRussell: It felt like this was a bit of an artificial structure, but maybe it's ok actually! I'll take the comment out.

PR #2, opened 2024-04-03
mrc-5191 Add tests

EmmaLRussell:
This branch adds a testing framework and some basic unit and e2e tests. 

Not really aiming for 100% test coverage at this point, as proof of concept code will change a lot, and we're on limited timescale for mvp.

For the unit tests, I don't really like the default scaffolded location of `__tests__` within the src folder - I prefer `tests` to be a sibling of `src` and to mirror its structure, so giving that a go here. 

I haven't done a 'data summary' for the Choropleth yet - the e2e tests just check that some geojson is rendered. I seemed to need to mock leaflet in the unit tests.. although that may be resolvable, because of the complexity of the leaflet set-up I think browser tests are going to be an easier way of testing the choropleth functionality, so the unit test here is very basic.


tests/unit/components/choropleth.spec.ts
```diff
@@ -0,0 +1,32 @@
+import { render, screen } from "@testing-library/vue";
+import { describe, expect, test } from "vitest";
+import { mockPinia } from "../mocks/mockPinia";
+import Choropleth from "../../../src/components/Choropleth.vue";
+
+const store = mockPinia();
+const renderComponent = async () => {
```
absternator: you dont need render to be async and await for it later

tests/unit/components/choropleth.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+                LGeoJson: true
+            }
+        }
+    });
+};
+
+describe("Choropleth", () => {
+    test("renders as expected", async () => {
+        await renderComponent();
+        const tileLayer = await screen.findByTestId("tile-layer");
```
absternator: no need to change but should be able to do getBy and not await...

tests/unit/layouts/default.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+    test("initialises data on load", async () => {
+        const appStore = useAppStore();
+        await renderLayout();
+        expect(appStore.initialiseData).toHaveBeenCalledTimes(1);
+    });
+
+    test("menu link navigates to About page", async () => {
+        await renderLayout();
+        const link = await screen.findByRole("button", { name: /About/i });
+        await fireEvent.click(link);
```
absternator: best to  use  `userEvent` from @testing-library/user-event

tests/unit/mocks/handlers.ts
```diff
@@ -0,0 +1,28 @@
...
+    ),
+    http.get("resources/geojson/geoBoundaries-TZA-ADM1_simplified.geojson", () =>
+        HttpResponse.json(MOCK_ADMIN1_GEOJSON.TZA)
+    ),
+    http.get("resources/geojson/geoBoundaries-TZA-ADM2_simplified.geojson", () =>
+        HttpResponse.json(MOCK_ADMIN2_GEOJSON.TZA)
+    ),
+    http.get("resources/geojson/geoBoundaries-MWI-ADM2_simplified.geojson", () =>
+        HttpResponse.json(MOCK_ADMIN2_GEOJSON.MWI)
+    )
```
absternator, line 28: add some  default handlers so it can catch any missed requests `const defaultHandlers = [http.all("*", () => new HttpResponse(null, { status: 200 }))];`

absternator (approved):
Awesome work tests look good.. Just left a basic comments :smile:

tests/unit/layouts/default.spec.ts
EmmaLRussell: Ah right, yeah, just read up on the difference - `userEvent` gives more realistic sets of events, not just the particular one you ask for.

tests/unit/components/choropleth.spec.ts
EmmaLRussell: Yeah, because in this case I don't actually need to wait for the data to be available (for loading to be false..)

PR #3, opened 2024-04-10
Add script for processing data

richfitz:
Add little script to process ADM1 data to json.

This uses the current version from Claire's files, and I've not committed the resulting files (though can do once we're happy with the format and path).


scripts/process-data.R
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env Rscript
+process_country <- function(x) {
+  id <- sprintf("%s-%s-%s", x$IS0, x$ID_0, trimws(x$ID_1))
```
EmmaLRussell:
```suggestion
  id <- sprintf("%s.%s_1", x$IS0, trimws(x$ID_1))
```

EmmaLRussell (commented):
This works great thanks. I think the ID generation needs tweaking a little. Talked this over with Wes yesterday and we're going to try to use the GID_1 and GID_2 IDs from the GADM data,. As far as I can tell, GID1 made up of the  ISO, the region id and the level (but not the numeric id of the country, ie ID_0). But we can always fiddle with that later if required.

richfitz:
Do you want me to commit the generated files too?

EmmaLRussell:
> Do you want me to commit the generated files too?

No that's alright, I think I'll just add them in a branch together with  Wes's geo files and code tweaks (hopefully just tweaks!) to get them all working together.

EmmaLRussell (approved):

PR #4, opened 2024-04-12
mrc-5205 Loading spinner

EmmaLRussell:
Adds a spinner to show over the map when data is loading. This should appear on initial page load, and the first time each country is selected (by clicking on it). 

I didn't really like the vuetify spinner, so using spin.js.

Also took out some over-cautious prevention of showing features on map while loading data - it was a bit alarming when everything disappeared while loading new regions!


src/composables/useLoadingSpinner.ts
```diff
@@ -0,0 +1,37 @@
...
+        } else if (spinner.el) {
+            // hide the spinner element so map can be clicked
+            spinner.el.hidden = true;
+            spinner.stop();
+        }
+    };
+
+    watch([spin, target], updateSpinner);
+
+    updateSpinner();
```
absternator, line 34: should we be calling this in a composable? or should it be returned and the component using this can use?

absternator (approved):
Nice work LGTM and works as expected :smile:

src/composables/useLoadingSpinner.ts
EmmaLRussell: Hmm, yeah, debatable. I think it's ok to do this here since the watch is built into the composable, so it makes sense for the spinner initialisation to happen here too.

PR #5, opened 2024-04-15
mrc-5196 Routing

EmmaLRussell:
This branch implements the requirement that there should be fixed routes to select indicator and country in the map. 

I've also included a base url which incorporates the currently hardcoded pathogen (dengue) and data version (may24) - future iterations will allow these to vary. 

When user browses to BASE_URL (e.g. http://localhost:3000 if you're running locally), they will be directed to:
BASE_URL/dengue/may24/FOI - where FOI is selected because it is the first indicator in the config. 
User can browse to  BASE_URL/dengue/may24/INDICATOR to select that indicator, with no country selected
User can browse to BASE_URL/dengue/may24/INDICATOR/COUNTRY_ISO to select that indicator and that country. 

Routes are case-insensitive. Back and forward buttons should work.

A future ticket will deal with setting selected country admin level through the route too. 

The code has been refactored so that when user clicks to select or deselect country, or clicks button to select indicator, it navigates to the relevant route rather than doing the selection directly through the store. So users can easily copy and share routes to data they browsed to.  These selections are done by the index page by responding to `indicator` and `country` props provided by the router. 

I've removed the unplugin auto router stuff as this was getting in the way of the custom routing. It wasn't really saving us much - it just means that we need to remember to manually add any new pages (like Privacy or Accessibility) to the router.  

There is a `notFound` page which we route to for any route which does not match any expected patterns - I also wanted to route to this if the requested indicator or country were not found. However this seemed tricky. We could push or replace a route to e.g. "/notFound", either in index page component, or in a route guard, but that would not retain the failed route in the browser address bar, which I think is preferable. Or, rather than using route params, we could generate routes for every combination of actual indicators and countries (which would be insane), or we could I suppose have added an aliased route on the fly to the router for the offending requested route - but that doesn't seem very nice either. So I've ended up handing nonexistent countries or indicators by rendering the notFound component from the index page, so the route is retained and we can also include a bit of detail about what the problem is. Let me know if you think of a better idea!

This branch has been deployed to the github pages location at https://mrc-ide.github.io/arbomap in order to test it with a non-root location.
e.g try browsing direct to:
https://mrc-ide.github.io/arbomap/dengue/may24/p9/ - to see p9 indicator selected
https://mrc-ide.github.io/arbomap/dengue/may24/FOI/TZA - to see FOI indicator and Tanzania selected

There are a couple of codecov grumbles - I'll skip the Choropleth ones since those should be covered by e2e tests, will do at the others once initial reviews are in.


src/components/Choropleth.vue
```diff
@@ -106,7 +123,10 @@ const createTooltips = {
         layer.bindTooltip(tooltipForFeature(feature)).openTooltip();
         layer.on({
             click: async () => {
-                await selectCountry(feature.properties[FEATURE_COUNTRY_PROP]);
+                const country = feature.properties[FEATURE_COUNTRY_PROP];
+                // select country, or unselect if click on it when already selected
+                const selectCountry = country === selectedCountryId.value ? "" : country;
```
absternator: sorry this ternary very confusing.. if you need to comment maybe best to split into 2 lines

src/pages/index.vue
```diff
@@ -1,27 +1,88 @@
...
+        (unknownCountry.value ? `Unknown country ISO: ${props.country}.` : "")
+    );
+});
+
+const selectDataForRoute = async () => {
+    if (!appConfig.value) {
+        return;
+    }
+
+    if (appConfig.value) {
```
absternator: Can be removed as already doing inverse check

src/pages/index.vue
```diff
@@ -1,27 +1,88 @@
...
+            const pattern = new RegExp(`^${props.indicator}$`, "i");
+            const indicator = Object.keys(appConfig.value.indicators).find((i) => pattern.test(i));
+            unknownIndicator.value = !indicator;
+
+            const country = props.country ? props.country.toUpperCase() : "";
+            unknownCountry.value = country && !appConfig.value.countries.includes(country);
+
+            if (!unknownIndicator.value && !unknownCountry.value) {
+                selectedIndicator.value = indicator;
+                await selectCountry(country);
```
absternator: just noticed but we should try {} catch {} our promises because they can be rejected... that should apply for all cases being drilled down

src/styles/settings.scss
```diff
@@ -26,3 +26,7 @@
   text-decoration: none;
 }
 
+p {
```
absternator: oof what was reasoning behind this? never a big fan of these glbal styles as its hard for the next person to debug why their p tag margins are weird

absternator (commented):
Works like a charm!! Nice work just left some small comments

src/styles/settings.scss
david-mears-2: I'm adding 3 static pages to the site with identical styling to each other, so it would be helpful to define p and ul globally so I don't have to duplicate those styles across 3 templates.

src/components/Choropleth.vue
david-mears-2: It's confusing to have this variable called `selectCountry` when that sounds like a function name (from the verb) and there is a function used elsewhere with that name. Let's call it 'selectedCountry'?

src/pages/notFound.vue
```diff
@@ -0,0 +1,25 @@
+<template>
```
david-mears-2, line 1: There's some slightly odd behaviour for normal non-existent routes - by which I mean things like 
`https://mrc-ide.github.io/arbomap/malaria` as opposed to when the indicator or country is wrong. This redirects to `https://mrc-ide.github.io/arbomap/dengue/may24/arbomap/malaria` and we see:
```Sorry, this page does not exist!

Unknown indicator: arbomap. Unknown country ISO: malaria.
```

david-mears-2: This is tolerable but I thought I'd mention it

src/pages/notFound.vue
```diff
@@ -0,0 +1,25 @@
...
+    <div v-if="appConfig" class="not-found ma-4 text-body-1">
+        <p>Sorry, this page does not exist!</p>
+        <p v-if="detail">
+            {{ detail }}
+        </p>
+        <p>
+            Go to the <router-link to="/">{{ appConfig.title }} home page</router-link> instead.
```
david-mears-2: When I follow this link from `https://mrc-ide.github.io/arbomap/dengue/asdf/FOI`, the browser back button doesn't work, because it takes me back to the homepage `https://mrc-ide.github.io/arbomap/dengue/may24/` which then redirects to `https://mrc-ide.github.io/arbomap/dengue/may24/FOI`.

This is a problem for anyone using the `https://mrc-ide.github.io/arbomap/dengue/may24` as their entry-point to the site (not just for people seeing the not found page)

david-mears-2: It seems like there is a redirect happening which also populates the browser history.

src/styles/settings.scss
david-mears-2: That said, I'm not sure which p tags this change is targeted at; if it's something on the map as opposed to the about page then I'd prefer these to have a more specific selector.

tests/e2e/routing.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    test("browse to indicator loads expected data", async ({ page }) => {
+        await page.goto(`${BASE_URL}/p9`);
+        await expectIndexPage(page, "/p9", "p9", "", "interpolateBlues", 33, 0);
+    });
+
+    test("browse to indicator and country loads expected data", async ({ page }) => {
+        await page.goto(`${BASE_URL}/FOI/TZA`);
+        await expectIndexPage(page, "/FOI/TZA", "FOI", "TZA", "interpolateReds", 173, 170);
+    });
```
david-mears-2, line 54: So we don't want the functionality of 'browse to country' without specifying an indicator?

tests/unit/mocks/mockRouter.ts
```diff
@@ -0,0 +1,12 @@
+import { createRouter, createWebHistory } from "vue-router/auto";
+import index from "../../../src/pages/index.vue";
+import about from "../../../src/pages/about.vue";
+
+export const mockRouter = () =>
+    createRouter({
+        history: createWebHistory(),
+        routes: [
+            { path: "/", componenth: index },
```
david-mears-2: componenth?

david-mears-2: Should this have caused a test failure?

src/pages/notFound.vue
david-mears-2: What's the intended entry-point url to the site? If that one doesn't result in a loop in browser-based navigation, and you can only get into the loop by visiting a non-existent route, I'm not too worried.

david-mears-2:
Looks good; see above for a question about a redirect causing the back button not to have an effect.

src/components/Choropleth.vue
```diff
@@ -1,28 +1,32 @@
...
+                :options="createTooltips"
+                :options-style="
+                    () => {
+                        return { ...style, fillColor: f.colour };
+                    }
+                "
+            >
+            </LGeoJson>
+        </LMap>
+        <div style="visibility: hidden" class="choropleth-data-summary" v-bind="dataSummary"></div>
```
david-mears-2, line 20: I think you added this just for testing, right? So we don't want screen-readers to try to interpret it? In which case it needs 'display: none' or just the HTML 'hidden' attribute

https://a11y-guidelines.orange.com/en/articles/accessible-hiding/

david-mears-2: Wait I'm wrong, `visibility: hidden` works fine for screen readers (omits the element for all types of users)

src/pages/notFound.vue
EmmaLRussell: Hmm, yeah, good point. I'll refactor to treat the pathogen and version like indicator and country and let the index page check if we know about both. I think that should fix that issue.

src/components/Choropleth.vue
EmmaLRussell: I think the comment is valid to clarify expected behaviour on user action, but yeah I'll refactor the logic for clarity.

src/pages/index.vue
EmmaLRussell: Yeah, though something's gone badly wrong if that's the case. Will add a separte ticket for error handling.

src/styles/settings.scss
EmmaLRussell: Yeah, it should be a consistent style for the non-map pages. Will add to a class.

src/pages/notFound.vue
EmmaLRussell: So the entry point could simply be arbomap.org, or could be arbomap.org/dengue/may24, or could also be direct link to indicator or indicator + country. All of these should work currently (I think).. But arbomap/dengue would also be a valid request, and that wouldn't work currently. Again, I'll roll supporting that into the refactor.

tests/e2e/routing.spec.ts
EmmaLRussell: Nope!

tests/unit/mocks/mockRouter.ts
EmmaLRussell: Oh wow! Probably! Will check,

src/pages/index.vue
EmmaLRussell: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-5273/Error-handling

tests/unit/mocks/mockRouter.ts
EmmaLRussell: I've added a test that the default layout actually renders the root component..

tests/e2e/routing.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+    selectedIndicator: string,
+    selectedCountry: string,
+    scale: string,
+    featureCount: number,
+    selectedCountryFeatureCount: number
+) => {
+    await page.waitForURL(new RegExp(`${BASE_URL}${url}`));
+    await expect(await page.textContent("button.bg-blue")).toBe(selectedIndicator);
+    const summary = await page.locator(".choropleth-data-summary");
+    await expect(await summary).toHaveAttribute("selected-indicator", selectedIndicator);
```
EmmaLRussell, line 16: Using more forgiving form of this test, which doesn't require attribute value to be present immediately.

EmmaLRussell: ...ugh, and had further e2e test misery, where one test started consistently failing on Safari only, without any substantive code changes.. Need to test on a real Safari instance really, hopefully when Rich is back tomorrow can try with his sacrifical test Mac. Have temporarily removed the Safari e2e config, and made ticket to investigate and reinstate: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-5274/Investigate-Safari-e2e-tests-failures

EmmaLRussell:
Should be ready for re-review. There are a couple of, hopefully spurious, e2e issues still outstanding - see comments - but would be good to get this merged so as not to hold up the other pages tickets. Have made a ticket to address one, and will reinstate the commented out spinner test in next  ticket.

src/pages/notFound.vue
david-mears-2: My browser history gets trapped in the loop if I visit
https://mrc-ide.github.io/arbomap
or
https://mrc-ide.github.io/arbomap/dengue/may24
because these both cause a redirect which the back button can't escape from. This doesn't need to be fixed in this PR, it could be a new ticket to address later.

david-mears-2:
Just one comment from me still to be addressed: https://github.com/mrc-ide/arbomap/pull/5#discussion_r1574850371

absternator (approved):
LGMT now thanks for addressing comments :smile:

EmmaLRussell:
> Just one comment from me still to be addressed: [#5 (comment)](https://github.com/mrc-ide/arbomap/pull/5#discussion_r1574850371)

Oops! Thanks for spotting. Should be fixed now.

david-mears-2 (approved):
:+1:

PR #8, opened 2024-04-23
mrc-5166 Use full dataset

EmmaLRussell:
This branch adds a full dataset (still with provisional data values), with related script changes and code changes to improve performance. (Huge line diff is mostly data!)

This branch is deployed to https://mrc-ide.github.io/arbomap

Load time is still pretty slow! Especially for selecting Brazil, the country with the most admin2 areas (~5s).  And distressingly it takes about as long to zoom back out to unselected view as it does to zoom in. Did some logging to investigate this, and seems that a lot of it is spent in Leaflet updates.  
It should at least show the spinner all the time while it is updating now. 

I wonder if doing something counterintuitive like having two Leaflet maps for the zoomed and non-zoomed view, and just strategically updating the z order might actually be quicker to change.. Or maybe not removing old Geojson objects from the map, but changing their colour to transparent when they shouldn't be shown... but I think that can wait for another branch!

It looks a bit weird without the selected country outlined and labelled - to be implemented in mrc-5203 

## Key changes

Raw data changes (in `/data/raw`):
- new indicator xlsx files
- admin1 geojson, which is reshaped from an array to a dictionary for easier consumption by the app to improve load time. 

Script changes:
- changes to R script to generate a single admin1 indicator file rather than per country (again, to improve initial load). Also output a list of countries which can easily be copied to the app's config files. Also, use new `GID` columns, and output 0s for sd - we aren't going to display sd, but awaiting confirmation about whether they should be in excel download.
- `transform_adm1_geojson_to_dict.cjs` script to do the reshape described above. 

Resource changes  (in `/public`, data used directly by the app):
- Indicators: full admin1 (in a single file) and admin2 data, as generated by the `process-data.R` script
- Geojson: full simplified admin1 (in a single file) and admin2 data (wizarded into existence by Wes).  Also: there a few countries (e.g. Bahamas) which do not have admin2 regions. For these cases, we keep detailed admin1 geojson in the `admin1` folder, and load that when the user selects that country. Also, `adm_0_bounds.json` which contains bounding box data for countries and for the global area, which we use to zoom  when a country is selected.
- `config.json` changes:
    - full list of `countries`
    - list of `countriesWithoutAdmin2`
    - id and name properties in the geosjon (rather than hardcoding those)
    - colour scales as requested by Clare  

Code changes:
- Choropleth
    - uses precalculated bounding boxes to update the bounds of the map rather than calculating from the geojson every time 
    - generates a border colour for region which desaturates its internal colour - this looks a bit subtler than uniform grey for outlines
    - uses new `waitingForMapBounds` in the store to show spinner as well as the `loading` data flag. This was introduced to deal with long waits for Leaflet to update after all data had been loaded. The waiting flag is set in the index pages when it first selects the new country (and I needed to debounce the action to give Vue some breathing time to get the spinner to show), and is unset when Leaflet eventually emits its `update:bounds` event. This feels quite insecure, having the flag set and unset by such unrelated areas of the app, but it seemed to be the most effective and reliable approach, after many other attempts!
    - computed `featureWithColours` is now a dict rather than an array to improve find by id, with a further computed to get the features as array when required. 
    
- useColourScale
    - support regions where no corresponding indicator exists (for countries where we have only partial data, e.g. Australia) - return transparent colour in this case
   
- index page
    - wait until all admin1 data is loaded before selecting country, to avoid weird initial view of selected country with incorrect colours
    - set `waitingForMapBounds` flag as described above

- appStore and resource utils
    - load new formats of data (single files for global data, bounding boxes, config changes)
    - deal with selecting a country which has no admin2 regions - load detailed admin1 geojson, and re-use admin1 indicators


tests/unit/mocks/mockObjects.ts
```diff
@@ -12,12 +20,8 @@ export const MOCK_APP_CONFIG = {
...
 
 export const MOCK_ADMIN1_GEOJSON = {
-    MWI: {
-        features: [{ properties: { shapeID: "123", shapeName: "Test123", shapeGroup: "MWI" } }]
-    },
-    TZA: {
-        features: [{ properties: { shapeID: "789", shapeName: "Test789", shapeGroup: "TZA" } }]
-    }
+    MWI: [{ properties: { shapeID: "123", shapeName: "Test123", shapeGroup: "MWI" } }],
+    TZA: [{ properties: { shapeID: "789", shapeName: "Test789", shapeGroup: "TZA" } }]
```
david-mears-2, line 28: (a question that came up while trying to rebase)
Why are you removing the 'features' property for this and not the MOCK_ADMIN2_GEOJSON? (One of my tests relies on it)

src/composables/useColourScale.ts
```diff
@@ -53,7 +53,12 @@ export const useColourScale = (selectedIndicators: Ref<Dict<FeatureIndicatorValu
         return result;
```
david-mears-2, line 53: in indicatorExtremes, line 51, when we define the default range:
```ts
indicatorRange.range =
                indicatorRange.max !== indicatorRange.min // Avoid dividing by zero if only one value...
                    ? indicatorRange.max - indicatorRange.min
                    : 1;
```

Does it matter that a range of 1 is actually quite a large range when we're working with certain metrics? A typical range for FOI is on the order of 0.02 and can be smaller.

tests/unit/mocks/mockObjects.ts
EmmaLRussell: To cut down on startup time we're now reading the admin1 geojson and indicators in from single files rather than files per-country. Since we needed to reformat the data anyway to do that, we've just pulled in the feature lists rather than the whole country level metadata that we don't need.

src/composables/useColourScale.ts
EmmaLRussell: We should possibly include a 'default range' for indicators. But I think in practice this issue is very unlikely to come up since we'll basically never have only a single region to display. We inherited this bit of logic from HINT where it can happen, but it's possibly overkill here.

david-mears-2:
It looks like you're updating p9 to serop9?

There are a few more finds and replaces to do in useColourScale.spec.ts, mockObjects.ts, tests/unit/pages/index.spec.ts, appStore.spec.ts.

I register my confusion that these tests are not failing as a result

EmmaLRussell:
> It looks like you're updating p9 to serop9?
> 
> There are a few more finds and replaces to do in useColourScale.spec.ts, mockObjects.ts, tests/unit/pages/index.spec.ts, appStore.spec.ts.
> 
> I register my confusion that these tests are not failing as a result

I suspect those tests don't fail because they don't rely on the app config, but are independent, but will check!

src/pages/index.vue
```diff
@@ -106,6 +112,19 @@ watch(
...
+watch([countryToSelect, admin1Indicators], async () => {
+    // wait until admin1 loaded before selecting country
+    if (countryToSelect.value !== null && Object.keys(admin1Indicators.value).length) {
+        // set waiting flag so Choropleth can show spinner while Leaflet updates. Debounce the select country
+        // action so Vue can start showing spinner first.
+        waitingForMapBounds.value = true;
+        debounce(() => {
+            selectCountry(countryToSelect.value);
+            countryToSelect.value = null;
+        })();
```
david-mears-2, line 124: I don't follow what the issue is that we're routing around. Why doesn't the line `this.loading = true` (which is the first thing `selectCountry` does, unless we're un-selecting a country) take care of the spinner? -- it seems the update to `this.loading` would be executed before the change to selectedCountryId, which I take to be what triggers Leaflet updating?

```ts
    async selectCountry(countryId: string) {
            if (countryId === this.selectedCountryId) {
                return;
            }

            if (!countryId) {
                this.selectedCountryId = "";
                return;
            }

            this.loading = true;

            // Some countries do not have admin2 regions or data - if one of these is selected, we load
            // a more detailed geojson, and re-use its admin1 indicators as "admin2"
            const level = this.appConfig.countriesWithoutAdmin2.includes(countryId) ? 1 : 2;

            if (!(countryId in this.admin2Indicators)) {
                if (level === 1) {
                    this.admin2Indicators[countryId] = this.admin1Indicators[countryId];
                } else {
                    this.admin2Indicators[countryId] = await getIndicators(countryId, level);
                }
            }
            if (!(countryId in this.admin2Geojson)) {
                this.admin2Geojson[countryId] = await getGeojsonFeatures(countryId, level);
            }

            this.selectedCountryId = countryId;
            this.loading = false;
        }
```

src/components/Choropleth.vue
```diff
@@ -1,17 +1,17 @@
...
     <div>
-        <LMap ref="map" style="height: 100vh; width: 100%" @ready="updateBounds">
+        <LMap ref="map" style="height: 100vh; width: 100%" @update:bounds="boundsUpdated">
             <LTileLayer data-testid="tile-layer" v-bind="backgroundLayer"></LTileLayer>
             <LGeoJson
-                v-for="f in featuresWithColours"
+                v-for="f in featuresWithColoursArr"
                 ref="featureRefs"
                 :key="getFeatureId(f.feature)"
                 :data-testid="getFeatureId(f.feature)"
```
david-mears-2: No great value in memo-izing this?

src/components/Choropleth.vue
```diff
@@ -55,50 +52,84 @@ const backgroundLayer = {
...
+
+const showSpinner = computed(() => waitingForMapBounds.value || loading.value);
+useLoadingSpinner(map, showSpinner);
 
-useLoadingSpinner(map, loading);
 const { getColour } = useColourScale(selectedIndicators);
 
-const getFeatureId = (feature: Feature) => feature.properties![FEATURE_ID_PROP];
-const getFeatureName = (feature: Feature) => feature.properties![FEATURE_NAME_PROP];
+const featureProps = computed(() => appConfig.value?.geoJsonFeatureProperties);
```
david-mears-2: For readability, I would vote for using the full term featureProperties to avoid conflation with Vue component props.

src/composables/useColourScale.ts
david-mears-2: Maybe
```ts
indicatorRange.range = indicatorRange.max - indicatorRange.min
```
is better then, for concision. Doesn't need to happen in this branch though.

david-mears-2 (approved):

src/pages/index.vue
EmmaLRussell: Oh, that bit works fine! The trouble is that we load everything and subsequently set loading to false, so the spinner goes away, but Vue+leaflet are still busy sorting their lives out for another (potentially) several seconds before the UI updates. So I wanted to prolong the time that the spinner was shown until view really had finished updating, and the only reliable place I found to hook into that was Leaflet declaring its bounds to have updated...

EmmaLRussell: This should become a moot point once Mantra works his optimisation magic... 🤞

src/components/Choropleth.vue
EmmaLRussell: I guess I must have used that in a test, but yeah it's a lot to render for the sake of a test if you've got thousands of the things..

src/components/Choropleth.vue
EmmaLRussell: Yep, fair enough, will update that.

src/components/Choropleth.vue
david-mears-2: I tried memo-izing to see if it made a difference to performance for Brazil, but I didn't notice one (though I didn't actually measure the time)

EmmaLRussell: Sorry, I misunderstood your comment - yeah, I doubt it would make a difference since the array itself is changing. It might do if we kept one v-for just for the admin1 regions, and just updated the visibility of the selected country ones... I think what Mantra's attempting will probably be more effective (if a bit less readable) by bypassing vue entirely...

src/components/Choropleth.vue
```diff
@@ -55,50 +53,84 @@ const backgroundLayer = {
...
+
+const showSpinner = computed(() => waitingForMapBounds.value || loading.value);
+useLoadingSpinner(map, showSpinner);
 
-useLoadingSpinner(map, loading);
 const { getColour } = useColourScale(selectedIndicators);
 
-const getFeatureId = (feature: Feature) => feature.properties![FEATURE_ID_PROP];
-const getFeatureName = (feature: Feature) => feature.properties![FEATURE_NAME_PROP];
+const featureProperties = computed(() => appConfig.value?.geoJsonFeatureProperties);
```
M-Kusumgar: does this need to be a computed? i was assuming this doesnt change at all

.github/workflows/e2e.yml
```diff
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
+      - "mrc-*"
```
M-Kusumgar: these are here because you want to deploy to github pages from this branch too? testing is very fast on this, could just run these on everything too but completely up to you!

src/components/Choropleth.vue
```diff
@@ -55,50 +53,84 @@ const backgroundLayer = {
...
-    });
+    if (!selectedInd) return {};
+    const selectedCountry = selectedCountryId.value;
+    return Object.fromEntries(
+        selectedFeatures.value.map((feature) => {
+            return [
+                getFeatureId(feature),
+                {
+                    feature,
+                    colour: getColourForFeature(feature, selectedInd, selectedCountry)
```
M-Kusumgar: also dont think you need to pass in selectedIndicator, you can just use it in the function but that dioesnt matter too much
```suggestion
                    colour: getColourForFeature(feature, selectedInd)
```

src/pages/index.vue
M-Kusumgar: just reviewing this now, will probably be changing a little bit of this in the next PR! there is a debounce but elsewhere :laughing:

M-Kusumgar (commented):
Looks good to me! Couple of things i noted which will probably be separate issues:
* Clicking and dragging leaves tooltips on the map
* Legend numbers are sometimes hard to read, e.g. on "SEROP9" indicator when the numbers go over the dark green, maybe a translucent which background on the legend?
* Double clicking or mash clicking on the map leads to weird behaviour, probably should disable all clicks until loading is finished

.github/workflows/e2e.yml
EmmaLRussell: Yeah, I think we want e2e on everything, might as well

src/components/Choropleth.vue
EmmaLRussell: The app config gets loaded along with the other resources so technically it isn't available immediately on app load, though in practice we could probably assume it exists for most components and just use it as is on load. I'll check we've got guard on the Choro and elsewhere and add that in.

src/components/Choropleth.vue
EmmaLRussell: Oh yeah, not sure what happened there! 😆

EmmaLRussell:
>[M-Kusumgar](https://github.com/M-Kusumgar) [May 1, 2024](https://github.com/mrc-ide/arbomap/pull/8#discussion_r1586296358)
>just reviewing this now, will probably be changing a little bit of this in the next PR! there is a debounce but elsewhere 😆

Yeah, I assumed this would end up being abolished by your reforms...

EmmaLRussell:
> Looks good to me! Couple of things i noted which will probably be separate issues:
> 
>    * Clicking and dragging leaves tooltips on the map
> 
There's already a ticket, but added the point that dragging causes problems...
https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-5265/Fix-phantom-tooltips

>    * Legend numbers are sometimes hard to read, e.g. on "SEROP9" indicator when the numbers go over the dark green, maybe a translucent which background on the legend?
> 

https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-5311/Give-legend-a-background

>    * Double clicking or mash clicking on the map leads to weird behaviour, probably should disable all clicks until loading is finished
https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-5312/Disable-map-interaction-while-loading

M-Kusumgar (approved):

PR #10, opened 2024-04-26
mrc-5279 Add map reset button

david-mears-2:


tests/unit/components/choropleth.spec.ts
```diff
@@ -7,12 +7,7 @@ const store = mockPinia();
 const renderComponent = () => {
     return render(Choropleth, {
         global: {
-            plugins: [store],
-            stubs: {
-                LMap: true,
-                LTileLayer: true,
-                LGeoJson: true
-            }
```
david-mears-2, line 15: I removed these because they didn't seem to affect whether the rendered Choropleth component contained `<l-map-stub>` or not (these leaflet components are stubbed in testSetup.ts). If they are supposed to achieve something else besides, then let me know.

EmmaLRussell: No that's fine, I guess this was superseded by the setup code.

src/components/ResetMapButton.vue
```diff
@@ -0,0 +1,43 @@
...
+const router = useRouter();
+
+const resetView = () => {
+    let homePath = `/${APP_BASE_ROUTE}/${props.selectedIndicator}`;
+    let currentPath = router.currentRoute.value.fullPath;
+    homePath = homePath.replace(/\/+$/, "");
+    currentPath = currentPath.replace(/\/+$/, "");
+    if (currentPath === homePath) {
+        // TODO: Do this by zooming out rather than by reloading the page, for performance reasons
+        window.location.reload();
```
EmmaLRussell: The easiest way to do this may be to emit an event which triggers the Choropleth to call its updateBounds method. If you merge with main, that should now use the fixed bounding boxes to zoom out so should be speedy. Might as well do that in this branch I think.

src/components/ResetMapButton.vue
```diff
@@ -0,0 +1,43 @@
...
+    }
+});
+
+const router = useRouter();
+
+const resetView = () => {
+    let homePath = `/${APP_BASE_ROUTE}/${props.selectedIndicator}`;
+    let currentPath = router.currentRoute.value.fullPath;
+    homePath = homePath.replace(/\/+$/, "");
+    currentPath = currentPath.replace(/\/+$/, "");
```
EmmaLRussell: It should be reliable here to just check if selectedCountryId is set - would be a bit simpler than this.

src/components/ResetMapButton.vue
```diff
@@ -0,0 +1,43 @@
...
+    <div class="leaflet-control-zoom leaflet-bar">
+        <a
+            @click.prevent="resetView"
+            class="leaflet-control-zoom-in"
+            href="#"
+            title="Reset map"
+            aria-label="Reset map"
+            role="button"
+        >
+            <vue-feather type="home" size="20" style="margin-top: 4px; margin-right: 1px"></vue-feather>
```
EmmaLRussell: Why mix in feather icons when we're already using the md ones? The md home-outline icon seems fine, or do you not like the eaves? https://pictogrammers.com/library/mdi/icon/home-outline/

EmmaLRussell (commented):
Looks great! Couple of small suggestions. Thanks for removing stubbing cruft. 

Might be nice to add an e2e test to check expected navigation occurs.

src/components/ResetMapButton.vue
david-mears-2: Yeah I was away with the fairies when I made this decision!

src/components/ResetMapButton.vue
david-mears-2: The problem is that when you're zoomed in but haven't selected any country, the path hasn't changed and nor has the selectedCountryId. However, using the updateBounds method to zoom out should fix all this.

EmmaLRussell: No, but if selectedCountryId is something other than empty, you should be able to reliably navigate to the base url. if it's already empty, you just need to updateBounds. if you just updated bounds when not empty, you wouldn't return to the global view - which is what you want. Maybe that's what you're saying though!

.gitignore
```diff
@@ -29,3 +29,7 @@ coverage
 .RData
 .Ruserdata
 data/processed
+/test-results/
+/playwright-report/
+/blob-report/
+/playwright/.cache/
```
david-mears-2, line 35: These were all generated when I installed playwright's VSCode extension.

tests/e2e/index.spec.ts
```diff
@@ -66,4 +66,15 @@ test.describe("Index page", () => {
         await expect(await page.innerText(".leaflet-tooltip-pane")).toContain("Bengo");
         await expect(await page.innerText(".leaflet-tooltip-pane")).toContain("FOI: 0.0134");
     });
+
+    test("resetMapButton click navigates to home page from country URL", async ({ page }) => {
+        await page.waitForURL(/dengue\/may24\/FOI/i);
+        const allRegions = await page.locator(GEOJSON_SELECTOR);
+        await expect(await allRegions).toHaveCount(1833);
+        await page.goto("/dengue/may24/serop9/MWI");
+        await expect(await allRegions).toHaveCount(1833);
```
EmmaLRussell: This should be a higher value than for home view, now that MWI is selected. This will be passing because it's 1833 immediately I guess.

EmmaLRussell (commented):
thanks, looks good! Just one little query on the e2e test.

tests/e2e/index.spec.ts
david-mears-2: Well caught. Updated.

EmmaLRussell (approved):

PR #12, opened 2024-05-01
mrc-5207 Indicator menu

EmmaLRussell:
This branch implements a menu for selecting indicator, replacing the simple buttons. The menu items include human readable name and description, and also support "indicator groups", to allow for the age stratified hospitalisation indicators. Each top level menu item corresponds to an indicator group (configured in app config), which has a "main indicator" and optionally multiple "sub-indicators". The sub-indicators are shown as swipeable slide group items. 

This branch also includes, opportunistically:
- full data update, including all the hospitalisation indicators and associated config
- .htaccess file for deployment to the production apache server, which will route any unknown urls to the index page, for front end routing to pick up
- z-index fix for legend button appearing on top of navigation drawer
- minor updates to About page from Clare

There are a couple of wrinkles which I think it would be best to deal with in separate tickets to stop this branch getting any staler. These are:
- truncated text on mobile - I think it might be best to have a static page with full indicator info rather than trying to crowbar it onto the menu (https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-5348/)
- Using the "main indicator" text and description for the menu item of a group - so the 'All ages' hospitalisation name and description for the 'hospitalisation' group. It would be better if the group itself could optionally have name and description as you could use more appropriate text for the group context. (https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-5345)


david-mears-2:
I find the flash-on-click on the age selector quite distracting. It's also not conveying anything, because there is a flash regardless of whether the click is successful - here's a screen recording of what I mean. When there are no more ages to scroll too, the button still does a flash.
[Screencast from 09-05-24 14:27:30.webm](https://github.com/mrc-ide/arbomap/assets/60350599/c58776b2-220a-42d0-bbcc-c9bfb83039f9)

david-mears-2:
We were talking about how the horizontal scroll of the sub-indicators degrade on smaller screens - the main issue is that on a desktop, with a small-ish window size, it's unwieldy to navigate left and right because you can't scroll or click-and-drag. And also you can't see that there are more sub-indicator options than those displayed.

scripts/process-data.R
```diff
@@ -1,18 +1,39 @@
 #!/usr/bin/env Rscript
 process_country <- function(x, country_codes, level) {
   if (level == 1) {
-    id <- trimws(x$GID_1)
+    id <- trimws(x$ID_1)
```
M-Kusumgar, line 4: did these just change because the data the scientists gave us changed these properties?

david-mears-2:
> truncated text on mobile - I think it might be best to have a static page with full indicator info rather than trying to crowbar it onto the menu

Or tooltips? (Opening on 'tap')

tests/unit/components/indicatorMenu/colourScaleIcon.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+                indicator: "hosp_total"
+            },
+            global: {
+                plugins: [mockPinia()]
+            }
+        });
+
+        const svg = (await screen.findByRole("presentation")) as HTMLElement;
+        expect(svg.getAttribute("height")).toBe("24");
+        expect(svg.getAttribute("width")).toBe("24");
```
david-mears-2, line 21: These seem like brittle tests - what if we want the icon to be a different size?

M-Kusumgar (commented):
Looks great! Couple of additional things ive discovered:
* clicking in the empty black space does not select the indicator
* i found myself wanting to see how the map changes as i vary the age groups in hospital admissions, perhaps we either show the sub groups with the drop downs closed or not close the menu until they click off it?
* just adding this in here so its written somewhere, not too fussed about it but when you select and indicator that doesnt have sub groups then there is a border around the whole box vs when you have indicator with sub groups the border is around the subgroup itself, that might be better in some ways, no idea

scripts/process-data.R
EmmaLRussell: Yup!

tests/unit/components/indicatorMenu/indicatorMenu.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+    const renderComponent = async () => {
+        await render(IndicatorMenu, {
+            global: {
+                plugins: [mockVuetify, mockPinia(), router]
+            }
+        });
+    };
+
+    const openMenu = async () => {
+        const button = await screen.findByText("Force of infection");
```
david-mears-2: Slightly more reliable would be getting by the class `indicator-menu-activator`, because other things might have the same text.

src/components/indicatorMenu/IndicatorMenu.vue
```diff
@@ -0,0 +1,84 @@
...
+                            @click="slideGroupClicked($event)"
+                        >
+                            <IndicatorSlideGroupItem
+                                :indicator-id="id"
+                                :indicator-name="
+                                    appConfig.indicators[appConfig.indicatorGroups[index].mainIndicator].shortName
+                                "
+                            ></IndicatorSlideGroupItem>
+                            <IndicatorSlideGroupItem
+                                v-for="subId in appConfig.indicatorGroups[index].subIndicators"
```
david-mears-2, line 51: I couldn't (immediately) work out how 'all ages' is making its way into this property 'subIndicators', given that 'all ages' is the main indicator. Where in the code is the main indicator added to the list of sub indicators?

src/composables/useColourScale.ts
```diff
@@ -55,7 +55,7 @@ export const useColourScale = (selectedIndicators: Ref<Dict<FeatureIndicatorValu
 
     const getColour = (indicator: string, featureIndicators: FeatureIndicatorValues | undefined) => {
         // If indicators do not exist for this feature, return transparent
-        if (!featureIndicators) {
+        if (!featureIndicators || !featureIndicators[indicator]) {
```
david-mears-2, line 77: checking - if `featureIndicators` _is_ present and `featureIndicators[indicator]` _isn't_, we should return the 0-colour?

tests/e2e/index.spec.ts
```diff
@@ -69,10 +72,11 @@ test.describe("Index page", () => {
         const firstRegion = await getNthRegion(page, 1);
         await firstRegion.hover();
         await expect(await page.innerText(".leaflet-tooltip-pane")).toContain("Bengo");
-        await expect(await page.innerText(".leaflet-tooltip-pane")).toContain("Force of infection: 0.0134");
+        await expect(await page.innerText(".leaflet-tooltip-pane")).toContain("Force of infection: 0.0455");
```
david-mears-2, line 75: I guess the number changes are from the new xlsx data?

tests/unit/pages/index.spec.ts
```diff
@@ -36,24 +34,19 @@ describe("Index page", () => {
...
-        expect((indicatorButtons[0] as HTMLButtonElement).textContent).toBe("FOI");
-        expect((indicatorButtons[0] as HTMLButtonElement).classList).toContain("bg-blue");
-        expect((indicatorButtons[1] as HTMLButtonElement).textContent).toBe("serop9");
-        expect((indicatorButtons[1] as HTMLButtonElement).classList).toContain("bg-black");
+        expect(await screen.findByText("Force of infection")).toBeVisible();
         expect(await screen.findByTestId("choropleth")).toBeVisible();
     });
 
-    test("button click routes to selected indicator", async () => {
+    /* test("button click routes to selected indicator", async () => {
```
david-mears-2: Can we delete this test now?

EmmaLRussell:
> I find the flash-on-click on the age selector quite distracting. It's also not conveying anything, because there is a flash regardless of whether the click is successful - here's a screen recording of what I mean. When there are no more ages to scroll too, the button still does a flash. [Screencast from 09-05-24 14:27:30.webm](https://github.com/mrc-ide/arbomap/assets/60350599/c58776b2-220a-42d0-bbcc-c9bfb83039f9)

Agreed, will remove this effect

tests/e2e/index.spec.ts
EmmaLRussell: They are!

EmmaLRussell:
> Looks great! Couple of additional things ive discovered:
> 
>     * clicking in the empty black space does not select the indicator
> 

Will fix this

>     * i found myself wanting to see how the map changes as i vary the age groups in hospital admissions, perhaps we either show the sub groups with the drop downs closed or not close the menu until they click off it?
> 

I'll see how it seems with explicit click away to close. 

>     * just adding this in here so its written somewhere, not too fussed about it but when you select and indicator that doesnt have sub groups then there is a border around the whole box vs when you have indicator with sub groups the border is around the subgroup itself, that might be better in some ways, no idea

Yeah, we discussed this, going to leave as is for now.

tests/unit/components/indicatorMenu/indicatorMenu.spec.ts
EmmaLRussell: selecting by class is discouraged by the testing framework, but I'll put in some more tests to check it's what we're expecting..

src/components/indicatorMenu/IndicatorMenu.vue
EmmaLRussell: It's not added to the list of sub-indicators, we just include an IndicatorSlideGroupItem for the main indicator before the subIndicators (immediately above).

src/composables/useColourScale.ts
EmmaLRussell: Yeah, so in the case where there are indicators configured which aren't present for all or some features - shouldn't really happen, but....

EmmaLRussell:
I've amended all the UI issues raised, apart from the ones I've detailed as going into separate tickets, so the following should be working:
- no ripple effect on menu items
- primary colour for selected items
- alignment should be the same whether menu item selected or now
- slide group arrows should always be shown
- full width of menu item should be clickable to select
- menu does not close on click to select, only when click away

I've also opportunistically fixed the issue where countries without admin 2 regions (e.g PRI and JAM) were not showing values - this was just a problem with using correct feature props.

src/components/indicatorMenu/IndicatorMenu.vue
```diff
@@ -5,7 +5,12 @@
                 {{ appConfig.indicators[selectedIndicator].humanReadableName }}
             </p>
             <v-icon :icon="menuOpen ? 'mdi-chevron-down' : 'mdi-chevron-up'" end></v-icon>
-            <v-menu activator="parent" v-model="menuOpen">
+            <v-menu
+                activator="parent"
+                v-model="menuOpen"
+                :close-on-content-click="false"
+                :close-on-back="false"
+            >
```
david-mears-2: The menu no longer closes on click away. Is that deliberate? I seem to expect that interaction to work.

david-mears-2: It seems the click away to close does work only when you click away above the menu, but not when you click away to the right of the menu.

EmmaLRussell: Good spot! This was because the vuetify-generated overlay content div under the menu was invisibly hogging most of the screen width and not responding to the click event. Putting the max width on that rather than the v-list fixes the issue.

david-mears-2 (approved):

M-Kusumgar (approved):
this feels good! happy to approve this, i think the indicator menu staying open is good actually, feels like i can quickly scan through different age groups.

potential issue for future: the white end of the color on the geojson feels like it blends in with the background tile layer, wonder what we can do about this, ill have a think too

EmmaLRussell:
> potential issue for future: the white end of the color on the geojson feels like it blends in with the background tile layer, wonder what we can do about this, ill have a think too

Yeah, I've noticed this a couple of times, particularly when zoomed into areas with low hospitalisations, it looks like you might not have any data!

PR #13, opened 2024-05-01
Mrc 5304 Optimise app + fade other countries on selection

M-Kusumgar:
The main idea here is working with leaflet instead of the vue wrapper for it. Main changes include:
* Remove GeoJson components from template
* loading flag is removed from store, this is no longer needed, we set loading within the choropleth component when a feature is clicked and remove loading flag when the map has updated the bounds, this produces the same behaviour as far as I can see
* We no longer need `featuresWithColours` the colours are put straight into style in `geoJSON` class initialisation
* `updateMap` is the function that does things in base leaflet and combines a couple of different functions together
* We fade and desaturate countries that arent selected

~~Deployed to ghp: https://mrc-ide.github.io/arbomap~~ #17 is now deployed to ghp


src/components/Choropleth.vue
```diff
@@ -150,59 +94,54 @@ const tooltipForFeature = (feature: Feature) => {
...
-                if (country === selectedCountryId.value) {
-                    countryToSelect = "";
-                } else {
-                    countryToSelect = country;
-                }
-                router.push(`/${APP_BASE_ROUTE}/${selectedIndicator.value}/${countryToSelect}`);
-            }
-        });
-    }
+const createTooltips = (feature: Feature, layer: Layer) => {
```
EmmaLRussell: Maybe we should rename this method since it isn't just creating tooltips any more.

src/components/Choropleth.vue
```diff
@@ -150,59 +94,54 @@ const tooltipForFeature = (feature: Feature) => {
...
+    geoJsonLayer.value.remove();
+
+    // create new geojson and add to map
+    geoJsonLayer.value = geoJSON<FeatureProperties, Geometry>(newFeatures, {
+        style,
+        onEachFeature: createTooltips
+    }).addTo(map.value.leafletObject);
+
+    // update bounds
+    const country = selectedCountryId.value || "GLOBAL";
```
EmmaLRussell: Ha, I just suggested to @david-mears-2 that it might be useful to call updateBounds in implementing the map reset button... so it might be nice to keep that as its own method.,

tests/e2e/index.spec.ts
```diff
@@ -61,9 +64,20 @@ test.describe("Index page", () => {
...
         await expect(await page.innerText(".leaflet-tooltip-pane")).toContain("FOI: 0.0134");
     });
+
+    test("selecting country fades colours of other countries", async ({ page }) => {
+        const firstRegion = await getNthRegion(page, 1);
+        const colour = await firstRegion.getAttribute("fill");
+        const stroke = await firstRegion.getAttribute("stroke");
+        const secondRegion = await getNthRegion(page, 1000);
+        await secondRegion.click();
+        await page.waitForURL(/dengue\/may24\/FOI/i);
```
EmmaLRussell: Do we need this? It would match immediately?

EmmaLRussell (commented):
This is much snappier than it was - thanks!

> loading flag is removed from store, this is no longer needed,

Yeah, I guess that's ok if it's not causing issues. i think there was a point where 'loading' was also indicating that you hadn't yet loaded appConfig or selected an indicator, but possibly everything that needs those are checking directly now.. 

Couple of issues I've seen:

Logged error after browsed to https://mrc-ide.github.io/arbomap/dengue/may24/FOI/BRA 
image

And this, which is actually a pre-existing issue.... but before it was just showing blank  page, now shows inifinite spinner, no error logged:
- got to a non-existent country: https://mrc-ide.github.io/arbomap/dengue/may24/FO/bran
- ..then browse to home page from app title

So not your fault, but might be worth looking into as part of this branch if you'd like to!

tests/e2e/index.spec.ts
```diff
@@ -61,9 +64,20 @@ test.describe("Index page", () => {
...
         await firstRegion.hover();
         await expect(await page.innerText(".leaflet-tooltip-pane")).toContain("Bengo");
         await expect(await page.innerText(".leaflet-tooltip-pane")).toContain("FOI: 0.0134");
     });
+
+    test("selecting country fades colours of other countries", async ({ page }) => {
+        const firstRegion = await getNthRegion(page, 1);
+        const colour = await firstRegion.getAttribute("fill");
+        const stroke = await firstRegion.getAttribute("stroke");
+        const secondRegion = await getNthRegion(page, 1000);
```
david-mears-2: Doesn't this get the 1000th region, not the second?

david-mears-2 (approved):
Incredible work

src/components/Choropleth.vue
david-mears-2: Leave it to me on the tooltips branch. I'll split this function into a bit that comes from a useTooltips composable, and a bit that remains here.

M-Kusumgar:
> Logged error after browsed to https://mrc-ide.github.io/arbomap/dengue/may24/FOI/BRA image

This should (hopefully) be fixed in the next PR, it was a case of unpacking vue reactivity i believe

> And this, which is actually a pre-existing issue.... but before it was just showing blank page, now shows inifinite spinner, no error logged:
> 
> * got to a non-existent country: https://mrc-ide.github.io/arbomap/dengue/may24/FO/bran
> * ..then browse to home page from app title
> 
> So not your fault, but might be worth looking into as part of this branch if you'd like to!

I have looked into this as part of the next PR too! it was a one line change, just had to update map when it was ready really

src/components/Choropleth.vue
M-Kusumgar: something like `bindTooltipAndClickEvent`? thats what ive gone with for now

src/components/Choropleth.vue
M-Kusumgar: sure can just pull it out into a function, i am assuming this will need to be further extracted into a composable as part of the map reset PR for david to be able to use this in another component mainly because `bounds` is not in the store? or maybe we should put `bounds` in the store and david can just use the updateBounds function?

tests/e2e/index.spec.ts
M-Kusumgar: yes sorry this was wrong, fixed now

tests/e2e/index.spec.ts
M-Kusumgar: yes my bad! i wanted to do second region and them i realised idk how many admin1 features the first region even has, have gone for last region now, think that is safer!

src/components/Choropleth.vue
EmmaLRussell: ..or could be something more generic reflecting what leaflet is doing, like `configureGeojsonLayer`

src/components/Choropleth.vue
EmmaLRussell: We were just going to make the reset button emit an event and let the choro handle it with its own `updateBounds`

tests/e2e/index.spec.ts
EmmaLRussell: I thought you meant 'second' as in 'the second one we're looking at here'! :laughing:

src/components/Choropleth.vue

EmmaLRussell (approved):
LGTM

PR #14, opened 2024-05-02
mrc-5210 Add gadm attribution and make sure it is visible

EmmaLRussell:
Added an attribution to GADM for the boundaries, and while I was at it, also fixed up the css issue where attributions were falling off bottom of screen (https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-5206)


david-mears-2 (approved):
All good. Let's move the legend down now that the map height is sorted.

In `Legend.vue` edit the `bottom` property of `.legend-container`.

PR #16, opened 2024-05-03
mrc-5316 Dataset with values for all regions

EmmaLRussell:
This branch updates data with Wes's hack to include values for all regions, so we shouldn't see gaps any more.

Targeting the performance optimisation branch in order to test performance - seems ok to me!

This is an interim dataset, to be replaced with final figures next week.


M-Kusumgar (approved):
awesome thanks! just tested the performance on this and seems fine to me too, it isnt performing that much worse, maybe ~50-100ms slower for brazil max but not very noticable!

since this is a completely independent change, could we instead point this to `mrc-5203`? that is my current working copy

EmmaLRussell:
> awesome thanks! just tested the performance on this and seems fine to me too, it isnt performing that much worse, maybe ~50-100ms slower for brazil max but not very noticable!
> 
> since this is a completely independent change, could we instead point this to `mrc-5203`? that is my current working copy

Sure!

EmmaLRussell:
> > awesome thanks! just tested the performance on this and seems fine to me too, it isnt performing that much worse, maybe ~50-100ms slower for brazil max but not very noticable!
> > since this is a completely independent change, could we instead point this to `mrc-5203`? that is my current working copy
> 
> Sure!

Actually, we could just point this to main and merge once the optimisation branch is in - just about to look at that now...

PR #17, opened 2024-05-03
Mrc 5203 Outline for country regions + misc bug fixes

M-Kusumgar:
I have tried to include comments in the diff instead of in the PR description, i think local comments could be better for these sparse diff changes, let me know what you think!

General strategy is  get admin level 0 geojson, convert coordinates to polyline and plot the line

Deployed to ghp: https://mrc-ide.github.io/arbomap


src/components/Choropleth.vue
```diff
@@ -5,6 +5,7 @@
             ref="map"
             style="height: calc(100vh - 48px); width: 100%"
             @update:bounds="waitingForMapBounds = false"
+            @ready="() => updateMap(selectedFeatures)"
```
M-Kusumgar, line 8: when you navigate to unknown country and then click home page, the map was not loading, we just needed an update when the map is ready basically

src/components/Choropleth.vue
```diff
@@ -41,13 +42,14 @@ const backgroundLayer = {
     minZoom: 3
 };
 
-const map = ref<typeof LMap | null>(null);
+const map = shallowRef<typeof LMap | null>(null);
```
M-Kusumgar, line 48: we dont need this reference to update when stuff inside it changes, we just need a reference to the component so have used `shallowRef` here

src/components/Choropleth.vue
```diff
@@ -41,13 +42,14 @@ const backgroundLayer = {
...
 const bounds = ref<Bounds | null>(null);
-const geoJsonLayer = ref<GeoJSON<FeatureProperties, Geometry>>(geoJSON(undefined));
+const geoJsonLayer = shallowRef<GeoJSON<FeatureProperties, Geometry>>(geoJSON(undefined));
+const countryOutlineLayer = shallowRef<Polyline | null>(null);
```
M-Kusumgar, line 51: same with these layers, only need shallow refs because we just destroy them and replace them every update

src/components/Choropleth.vue
```diff
@@ -129,29 +131,43 @@ const style = (f: Feature) => {
 };
 
 const updateBounds = async () => {
-    // update bounds
+    const rawMap = toRaw(map.value);
```
M-Kusumgar: `toRaw` unpacks an object and returns it without vue reactivity attached to it, we are not using the map in a reactive way (to trigger changes for example) so this should be fine, i think this gets rid of the map is null errors that @EmmaLRussell showed here: https://github.com/mrc-ide/arbomap/pull/13#pullrequestreview-2035690703

src/stores/appStore.ts
```diff
@@ -100,6 +102,7 @@ export const useAppStore = defineStore("app", {
                 this.admin2Geojson[countryId] = await getGeojsonFeatures(countryId, level);
             }
 
+            this.admin0Geojson = (await getGeojsonFeatures(countryId, 0))[0];
```
M-Kusumgar: theres only one feature, the actual country so just unpacking the 0th index

src/components/Choropleth.vue
```diff
@@ -129,29 +131,43 @@ const style = (f: Feature) => {
...
 
     // create new geojson and add to map
     geoJsonLayer.value = geoJSON<FeatureProperties, Geometry>(newFeatures, {
         style,
         onEachFeature: configureGeojsonLayer
-    }).addTo(map.value.leafletObject);
+    }).addTo(leafletMap);
+
+    if (admin0Geojson.value) {
+        const latLngs = GeoJSON.coordsToLatLngs((admin0Geojson.value.geometry as any).coordinates, 2);
```
M-Kusumgar, line 163: nice utility to turn coordinates into a line

src/components/Choropleth.vue
```diff
@@ -129,29 +138,46 @@ const style = (f: Feature) => {
...
 
     // create new geojson and add to map
     geoJsonLayer.value = geoJSON<FeatureProperties, Geometry>(newFeatures, {
         style,
         onEachFeature: configureGeojsonLayer
-    }).addTo(map.value.leafletObject);
+    }).addTo(leafletMap);
+
+    if (admin0Geojson.value) {
+        const latLngs = GeoJSON.coordsToLatLngs(admin0Geojson.value.geometry.coordinates, 2);
```
M-Kusumgar: nice utility to turn coords into line

src/components/Choropleth.vue
```diff
@@ -129,29 +138,46 @@ const style = (f: Feature) => {
...
 
     // record bounds for testing
     bounds.value = { west, east, south, north };
 };
 
 const updateMap = async (newFeatures: Feature[]) => {
-    if (!map.value) return;
+    const rawMap = toRaw(map.value);
+    if (!rawMap) return;
+    const leafletMap = rawMap.leafletObject;
```
david-mears-2: These three lines recur twice, could they go into a `computed`?

src/components/Choropleth.vue
```diff
@@ -129,29 +138,46 @@ const style = (f: Feature) => {
...
+    if (admin0Geojson.value) {
+        const latLngs = GeoJSON.coordsToLatLngs(admin0Geojson.value.geometry.coordinates, 2);
+        // className is just for testing
+        countryOutlineLayer.value = polyline(latLngs, {
+            color: "black",
+            weight: 1,
+            opacity: 0.5,
+            className: "country-outline"
+        }).addTo(leafletMap);
+    }
```
david-mears-2: Just a readability nitpick - could you either extract this to a function `addCountryOutline`, or add a comment just before that says that these lines are adding the country outline?

src/stores/appStore.ts
```diff
@@ -27,7 +27,8 @@ export const useAppStore = defineStore("app", {
         admin1Indicators: {},
         admin1Geojson: {},
         admin2Indicators: {},
-        admin2Geojson: {}
+        admin2Geojson: {},
+        admin0Geojson: null
```
david-mears-2: ~Is there some reason why `admin1Geojson` and `admin2Geojson` are empty objects, while `admin0Geojson` is null?~

~Is it just so that you can write `if (admin0Geojson.value)` (because null is falsy and `{}` is truthy)?~

~I feel like the downside of implying that the property is of a different structure than its brethren is greater than the verbosity downside of writing `if (Object.keys(admin0Geojson.value) === 0)` or importing `isEmpty` from lodash: `if (isEmpty(admin0Geojson.value))`~

src/stores/appStore.ts
david-mears-2: Oh I see - so `admin0Geojson` will in fact be a different structure from `admin1Geojson` and `admin2Geojson`? 1 and 2 being key-value dicts, but admin0Geojson being a single 'value', the one feature, if I understood correctly?

I feel having different structures is likely to cause slight confusion. A solution could be renaming the property to something that does not imply a strong analogy to `adminNGeojson`, e.g renaming to `admin0GeojsonFeature` (or renaming the dicts to use a plural term `geojsons`). Or, enforcing the analogy to reduce surprise, so that `admin0Geojson` becomes a single-key dict. What do you think?

david-mears-2: Maybe I'm just not used to using languages that enforce you to document types - it would be reasonable to say that a confused reader should just refer to storeTypes.ts to find out the structures of the properties.

src/stores/appStore.ts
david-mears-2: ^ Wrote the above before I worked out that `admin0Geojson` had a different structure from the other `adminNGeojsons`: https://github.com/mrc-ide/arbomap/pull/17#discussion_r1589294467

tests/e2e/index.spec.ts
```diff
@@ -82,6 +84,14 @@ test.describe("Index page", () => {
         await expect(await firstRegion.getAttribute("stroke")).not.toEqual(stroke);
     });
 
+    test("country outline is shown", async ({ page }) => {
+        const firstRegion = await getNthRegion(page, 1);
+        await firstRegion.click();
+        await page.waitForURL(/dengue\/may24\/FOI\/AGO/i);
+        await expect(await page.locator("div.spinner")).toHaveCount(0);
+        await expect(await page.locator(COUNTRY_OUTLINE_SELECTOR)).toHaveCount(1);
+    });
```
david-mears-2: Checking that we start with no country outline might add extra confidence (that the country outline is only shown when needed):

```suggestion
    test("country outline is shown", async ({ page }) => {
        const firstRegion = await getNthRegion(page, 1);
        // assuming page is fully loaded
        await expect(await page.locator(COUNTRY_OUTLINE_SELECTOR)).toHaveCount(0);
        await firstRegion.click();
        await page.waitForURL(/dengue\/may24\/FOI\/AGO/i);
        await expect(await page.locator("div.spinner")).toHaveCount(0);
        await expect(await page.locator(COUNTRY_OUTLINE_SELECTOR)).toHaveCount(1);
    });
```

david-mears-2:
I really like the appearance of the saturated country against the desaturated background. 

What are the 'misc bug fixes' that you've tackled here?

src/components/Choropleth.vue
M-Kusumgar: unfortunately the map is already a ref in vue, I am unpacking vue reactivity here, so i can't move it into a computed because that just wraps in it vue reactivity again. I can however do something like
```typescript
const leafletMap = toRaw(map.value)?.leafletObject;
if (!leafletMap) return;
```
to reduce the line count :laughing:

tests/e2e/index.spec.ts
M-Kusumgar: yh nice!

src/components/Choropleth.vue
david-mears-2: Ah, I see. Can we use a basic non-computed function then?

david-mears-2: (I'm more concerned with avoiding duplication than with reducing line count)

tests/e2e/index.spec.ts
david-mears-2: I shouldn't have used the 'suggest' feature here, because the line `// assuming page is fully loaded` is just a placeholder for somebody actually checking that the page has loaded by this point in the test. (Or we risk false positives)

src/components/Choropleth.vue
M-Kusumgar: ive added a comment on top if the if statement

M-Kusumgar:
> I really like the appearance of the saturated country against the desaturated background.

 yh im really digging the look too!

> What are the 'misc bug fixes' that you've tackled here?

so there are 2 fixes:
* if you go to a country that doesnt exist via url (e.g. `../FOI/BRAN`) and then click go back to home page, then in the current main version you see an infinite spinner and the map doesnt load. This [comment](https://github.com/mrc-ide/arbomap/pull/17/files#r1589052375) is attached to the line that fixes that issue
* there were console errors about the `this._map` variable being null from leaflet, i had been getting those at random intervals too, emma's [comment](https://github.com/mrc-ide/arbomap/pull/13#pullrequestreview-2035690703) has both these issues in a bit more detail but after looking online this seems to be an issue with vue's reactivity, so that is why I have used `toRaw` function in a couple places

src/stores/appStore.ts
M-Kusumgar: i could have gone for an object (or actually array here because the admin level 0 geojsons are a little different) however i feel like nulls are better than empty objects/arrays as initialisers because you can do falsy checks on them, you cant do those for empty objects and empty arrays

but this is all just preference!

src/components/Choropleth.vue
EmmaLRussell: I don't quite understand what that's giving us that just checking if value was falsy wouldn't...? Could you put a comment here about what problem this is solving, because this will probably seem mysterious in a couple of months time!

src/components/Choropleth.vue
EmmaLRussell: So this isn't a GeoJson layer as such, but a polyline instead?

EmmaLRussell (approved):
Looks great to me, just a suggestion for adding a comment into the code.

src/stores/appStore.ts
M-Kusumgar: this is a fair point actually, will change the name to `admin0GeojsonFeature` instead!

src/components/Choropleth.vue
M-Kusumgar: ohhh i got you, yes i can extract this into a function!

tests/e2e/index.spec.ts
M-Kusumgar: ohhh okay got you, i can fill that in!

src/components/Choropleth.vue
M-Kusumgar: ~~i found the original link (also put that in the comment) and have tried to explain it in my own words in terms of what i think is happening with the vue reactivity!~~

M-Kusumgar: in putting the explanation online into my own words i realised that the `shallowRef` stuff is sufficient actually! basically, vue 3 deep watches refs by default but that means that it updates every time the leaflet map object updates which has its own reactivity built into it. So we lose the reference to `map` sometimes when we try and do an update to `map` because vue 3 wants to react to the changes to the internals of `map`

david-mears-2 (approved):

PR #7, opened 2024-04-19
mrc-5208 Format tooltips and extract to composable

david-mears-2:
~~This branch builds on the mrc-5196 branch and should be kept rebased on that branch to avoid future git conflicts.~~

This branch builds on the mrc-5166 branch and should be kept rebased on that branch to avoid future git conflicts.

- [x] Should not be merged until #5 is merged.
- [ ] Keep up-to-date with #8.

PR #18, opened 2024-05-03
Mrc 5208: Format tooltips and extract to composable (v3)

david-mears-2:
Rebased onto main for your delectation.

The original PR with review comments is here: https://github.com/mrc-ide/arbomap/pull/7

I’m hoping to be able to merge this one in before main diverges away again!

### DIFF
--- a/src/composables/useTooltips.ts
+++ b/src/composables/useTooltips.ts
@@ -1,0 +1,44 @@
+import { storeToRefs } from "pinia";
+import { useAppStore } from "../stores/appStore";
+
+export const useTooltips = () => {
+    const { selectedIndicators, selectedIndicator, appConfig } = storeToRefs(useAppStore());
+
+    const sortedIndicators = computed(() => {
+        const sortedKeys = Object.keys(appConfig.value.indicators).sort((indicatorName) =>
+            indicatorName.toLowerCase() === selectedIndicator.value.toLowerCase() ? -1 : 1
+        );
+
+        const sortedMap = new Map();
+
+        sortedKeys.forEach((key) => {
+            sortedMap.set(key, appConfig.value.indicators[key]);
+        });
+
+        return sortedMap;
+    });
+
+    const tooltipForFeature = (featureId: string, featureName: string) => {
+        const tooltipOptions = { sticky: true };
+        if (!(featureId in selectedIndicators.value)) return null;
+        const featureValues = selectedIndicators.value[featureId];
+        let indicatorValues = "";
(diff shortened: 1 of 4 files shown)


david-mears-2:
I have split out adapting the tooltips for larger number of indicators into [its own ticket](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-5322/Adapt-tooltips-for-larger-number-of-indicators), because I'd really like to get the basic version merged into main first (the branch has needed repeated maintenance to keep it up to date).

M-Kusumgar (approved):
Looking nice, much more readable, what was the sticky option for?

david-mears-2:
> Looking nice, much more readable, what was the sticky option for?

Thanks. I should have said, it makes the tooltip follow the cursor. https://leafletjs.com/reference.html#tooltip-sticky

I think this is just an aesthetic preference of mine, so feel free to disagree.

https://github.com/mrc-ide/arbomap/assets/60350599/86fd776c-f4b1-4e3e-ad94-1d71400c0617

src/composables/useTooltips.ts
```diff
@@ -0,0 +1,44 @@
+import { storeToRefs } from "pinia";
+import { useAppStore } from "../stores/appStore";
+
+export const useTooltips = () => {
+    const { selectedIndicators, selectedIndicator, appConfig } = storeToRefs(useAppStore());
+
+    const sortedIndicators = computed(() => {
+        const sortedKeys = Object.keys(appConfig.value.indicators).sort((indicatorName) =>
+            indicatorName.toLowerCase() === selectedIndicator.value.toLowerCase() ? -1 : 1
+        );
```
EmmaLRussell: This is sorting on the indicator id, but it would be better to sort on human readable name since that's what we're displaying.

EmmaLRussell (commented):
Looks great, just one suggestion on the sorting.

src/composables/useTooltips.ts
david-mears-2: Oh yes, I need to make sure that the ids match, not compare a name with an id.

david-mears-2: In fact, this code was correctly comparing ids/keys with ids/keys (no names involved); so I'm just renaming the param from 'indicatorName' to 'indicatorId'.

david-mears-2: (NB this isn't an alphabetical sort, it's sorting the selected indicator to the top of the list.)

PR #19, opened 2024-05-03
Mrc 5169 Lock on to country bounds and set min zoom + remove empty scrollbar

M-Kusumgar:
This PR should lock the bounds and minimum zoom of the leaflet map view. There was also an annoying empty scrollbar on the side of the map. Just got rid of that with some css :)

Main changes:
* flag for when country is selected, this is to stop us from locking the bounds every time someone changes anything, otherwise the user would only be able to zoom in and never zoom out
* I reset the `maxBounds` and `minZoom` of the leaflet map every time the features change to make sure we can actually zoom out all the way when we go to global view

Note: you may be thinking we have the bounding boxes, why not just set the bounds with that, the bounding boxes are too tight and leaflet already does its own calculation about fitting the bounding box with the maximum zoom out possible, so we let it do its thing, once it is done we just use those bounds and lock them in

Deployed to ghp: https://mrc-ide.github.io/arbomap

The tooltips will stay when you click and drag on locked boundaries, this will be fixed in the next PR!

### DIFF
--- a/src/components/Choropleth.vue
+++ b/src/components/Choropleth.vue
@@ -4,7 +4,7 @@
             class="map"
             ref="map"
             style="height: calc(100vh - 48px); width: 100%"
-            @update:bounds="lockBounds"
+            @update:bounds="mapBoundsUpdated"
             @ready="() => updateMap(selectedFeatures)"
             :max-bounds-viscosity="1"
         >
@@ -145,7 +145,7 @@ const style = (f: Feature) => {
 
 const getLeafletMap = () => map.value?.leafletObject;
 
-const lockBounds = async () => {
+const mapBoundsUpdated = async () => {
     const leafletMap = getLeafletMap();
     if (!leafletMap) return;
 
@@ -158,17 +158,17 @@ const lockBounds = async () => {
     waitingForMapBounds.value = false;
 };
 
-const getBoundsForCountry = (countryId?: string) => {
+const getRegionBounds = (countryId?: string) => {
     if (Object.keys(countryBoundingBoxes.value).length === 0) return null;
     const [west, east, south, north] = countryBoundingBoxes.value[countryId || "GLOBAL"];
     return new LatLngBounds([south, west], [north, east]);
 };
 
 const updateBounds = async () => {
-    const leafletMap = toRaw(map.value)?.leafletObject;
+    const leafletMap = getLeafletMap();
     if (!leafletMap) return;
 
-    const countryBounds = getBoundsForCountry(selectedCountryId.value);
+    const countryBounds = getRegionBounds(selectedCountryId.value);
     if (!countryBounds) return;
 
     bounds.value = countryBounds;
@@ -176,13 +176,13 @@ const updateBounds = async () => {
 };
 
 const resetMaxBoundsAndZoom = () => {
-    const leafletMap = toRaw(map.value)?.leafletObject;
+    const leafletMap = getLeafletMap();
     if (!leafletMap) return;
 
-    const globalBounds = getBoundsForCountry();
+    const globalBounds = getRegionBounds();
     if (!globalBounds) return;
 
-    leafletMap.setMinZoom(2.5);
+    leafletMap.setMinZoom(3);
     leafletMap.setMaxBounds(globalBounds);
 };
 
--- a/src/components/Choropleth.vue
+++ b/src/components/Choropleth.vue
@@ -157,24 +157,34 @@ const lockBounds = async () => {
 };
 
 const getBoundsForCountry = (countryId?: string) => {
+    if (Object.keys(countryBoundingBoxes.value).length === 0) return null;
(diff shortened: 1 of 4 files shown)


src/styles/settings.scss
```diff
@@ -9,6 +9,10 @@
 //   $color-pack: false
 // );
 
+html {
+  overflow-y: auto;
+}
+
```
M-Kusumgar, line 15: this is for getting rid of that pesky empty scrollbar

david-mears-2:
Don't know if you want to fix this here or in a separate ticket, but when you zoom out from the world view, it looks weird (everything disappears except the coloured indications). We could tighten the bounds to prevent zooming out.
[Screencast from 07-05-24 18:15:28.webm](https://github.com/mrc-ide/arbomap/assets/60350599/e2105673-7acb-40eb-9c7c-006782719295)

src/components/Choropleth.vue
```diff
@@ -142,20 +145,48 @@ const style = (f: Feature) => {
...
+    }
 
-    // record bounds for testing
-    bounds.value = { west, east, south, north };
+    isNewSelectedCountry.value = false;
+    waitingForMapBounds.value = false;
 };
 
-const updateMap = async (newFeatures: Feature[]) => {
+const getBoundsForCountry = (countryId?: string) => {
```
david-mears-2:
```
const getBounds = (countryId?: string = "GLOBAL") => {
```
Maybe rename because this function doesn't always deal with countries. `getBounds` or `getRegionBounds`?

david-mears-2 (approved):
I left two comments, but it looks good.

src/components/Choropleth.vue
```diff
@@ -142,20 +145,48 @@ const style = (f: Feature) => {
 
 const getLeafletMap = () => map.value?.leafletObject;
 
-const updateBounds = async () => {
+const lockBounds = async () => {
```
EmmaLRussell: Could we call this `mapBoundsUpdated` rather than `lockBounds`, since it doesn't always actually lock the bounds, but it is called in respons to the update bounds event.

src/components/Choropleth.vue
```diff
@@ -142,20 +145,48 @@ const style = (f: Feature) => {
...
+
+    const countryBounds = getBoundsForCountry(selectedCountryId.value);
+    if (!countryBounds) return;
+
+    bounds.value = countryBounds;
+    await leafletMap.fitBounds(bounds.value);
+};
+
+const resetMaxBoundsAndZoom = () => {
+    const leafletMap = toRaw(map.value)?.leafletObject;
```
EmmaLRussell: I thought you'd decided you didn't need `toRaw`?

src/components/Choropleth.vue
EmmaLRussell: I'd vote for `getRegionBounds`

EmmaLRussell (commented):
Nice! Works great.

It's quite a confusing little interlocking set of moving parts this, especially how the isNewSelectedCountry flag gets set very seprately from the updateBounds method which actually sets the bounding box bounds into the map (triggered by watch on selectedCountry, and, via updateMap, on selectedFeatures respectively), but then they come together in the lock method when leaflet notifies about having updated its bounds... So I think I'd vote for putting a couple of comments in the code to help out our future selves!

M-Kusumgar:
> Don't know if you want to fix this here or in a separate ticket, but when you zoom out from the world view, it looks weird (everything disappears except the coloured indications). We could tighten the bounds to prevent zooming out. [Screencast from 07-05-24 18:15:28.webm](https://github.com/mrc-ide/arbomap/assets/60350599/e2105673-7acb-40eb-9c7c-006782719295)

this should be fixed now! yh the min zoom was a little too low thats all

src/components/Choropleth.vue

src/components/Choropleth.vue

src/components/Choropleth.vue
M-Kusumgar: ooops that was a merge conflict mistake!

M-Kusumgar:
> Nice! Works great.
> 
> It's quite a confusing little interlocking set of moving parts this, especially how the isNewSelectedCountry flag gets set very seprately from the updateBounds method which actually sets the bounding box bounds into the map (triggered by watch on selectedCountry, and, via updateMap, on selectedFeatures respectively), but then they come together in the lock method when leaflet notifies about having updated its bounds... So I think I'd vote for putting a couple of comments in the code to help out our future selves!

yh actually ive been thinking about this quite a bit for the admin 1 and admin 2 toggle actually, it has gotten a bit more complicated with more ifs, wondering how to make this cleaner and make more sense, will have a think today about this

edit: as discussed I have made a ticket to extract out leaflet into a composable and will put comments and stuff on there, ticket: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-5335/Extract-Leaflet-logic-into-a-composable

EmmaLRussell (approved):

PR #20, opened 2024-05-03
Mrc 5265 Fix phantom tooltips

M-Kusumgar:
Phantom tooltips are a real issue especially with locked bounds, clicking and dragging opened loads of tooltips, this PR fixes it by tracking the layer whose tooltip is open and closing the tooltip of any previous layer. This is not optimal but it didnt seem like there was much better online about this as far as I could tell.

Unfortunately I cannot seem to recreate this bug on playwright which is quite annoying. I did the standard `dragTo` or even the manual way with mouse down, hover and mouse up but i could not get the screenshot correct in playwright. Since this is a bug fix I left the testing for that but am open to any suggestions!

### DIFF
--- a/.github/workflows/deploy-to-ghp.yml
+++ b/.github/workflows/deploy-to-ghp.yml
@@ -4,7 +4,7 @@ name: Deploy static content to Pages
 on:
   # Runs on pushes targeting the default branch
   push:
-    branches: ['main', 'mrc-5203']
+    branches: ['main', 'mrc-5169']
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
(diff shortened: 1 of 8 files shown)


david-mears-2 (approved):
Lovely elegant solution

EmmaLRussell (commented):
Great stuff. I wonder if this rightly belongs in the new `useTooltips` composable? (Not currenlty merged into this branch.)

Again, just because I have a memory like a sinkhole, I'd probably pop  a little comment on there about the issue those lines are addressing.

M-Kusumgar:
> Great stuff. I wonder if this rightly belongs in the new `useTooltips` composable? (Not currenlty merged into this branch.)

the fix for this PR is in the `configureGeojsonLayer` instead of strictly tooltip related, i think this should go in the leaflet composable instead maybe

> Again, just because I have a memory like a sinkhole, I'd probably pop a little comment on there about the issue those lines are addressing.

have left a comment!

EmmaLRussell (approved):

PR #21, opened 2024-05-08
Mrc 5335 Extract leaflet into composable + clean up choropleth

M-Kusumgar:
There were a lot of moving parts in the choropleth component and it was getting too complicated so I have tried to make things as neat as I could in this PR by extracting things into composables, any further suggestions are much appreciated! For testing of the `useLeaflet` we will continue to rely on e2e tests.

Main changes:
* `fadeColour` is now in `useColorScales` i think this makes sense
* `backgroundLayer` and corresponding global `minZoom` are extracted into utils since these dont change
* added `useDataSummary` composable since this is completely independent from the functionality, i have not tested this since it is used in the e2e tests, feels weird to test something that is just used for testing idk
* Extracted leaflet logic into `useLeaflet` composable, only functionally different thing ive added is `smoothFactor: 0` otherwise should be the same


src/composables/useColourScale.ts
```diff
@@ -53,35 +59,53 @@ export const useColourScale = (selectedIndicators: Ref<Dict<FeatureIndicatorValu
...
+                color: "transparent"
+            };
         }
 
-        const value = featureIndicators[indicator].mean;
-
         if (!indicatorExtremes.value || !colourScales.value) {
-            return "rgb(200, 200, 200)";
+            const colour = "rgb(200, 200, 200)";
+            const baseColour = isFaded ? fadeColour(colour) : colour;
```
EmmaLRussell: Feels like we could define this 'missing' colour and its faded equivalent once rather than on every feature we need it..

src/utils.ts
```diff
@@ -9,3 +9,14 @@ export function debounce(fn, wait = 0) {
...
+export const minZoom = 3;
+
+export const backgroundLayer = {
+    url: "https://server.arcgisonline.com/ArcGIS/rest/services/Canvas/World_Light_Gray_Base/MapServer/tile/{z}/{y}/{x}",
+    attribution:
+        "Tiles &copy; Esri &mdash; Esri, DeLorme, NAVTEQ; " +
+        "Boundaries: <a href='https://gadm.org' target='_blank'>GADM</a> version 4.1",
+    maxZoom: 10,
+    minZoom
+};
```
EmmaLRussell: I get that it makes the Choro module a bit tidier to pull these out, but since they're only ever going to be used to Chorpleth, wouldn't mind them going into `composables/utils` or even `choropleth/map/utils` then could potentially tidy Legend and ResetMapButton into the `map` folder too.

src/components/Choropleth.vue
```diff
@@ -13,212 +13,87 @@
...
-}));
+const getTooltip = (feature: Feature) => tooltipForFeature(getFeatureId(feature), getFeatureName(feature));
 
-const configureGeojsonLayer = (feature: Feature, layer: Layer) => {
-    const tooltip = tooltipForFeature(getFeatureId(feature), getFeatureName(feature));
-    if (tooltip) layer.bindTooltip(tooltip.content, tooltip.options);
-    layer.on({
-        click: async () => {
-            waitingForMapBounds.value = true;
+const layerOnEvents = (feature: Feature) => {
```
EmmaLRussell, line 71: I can see why you've kept this out of the useLeaflet composable, but it does feel a little bit backwards to passing this in, so I think a comment here about the context in which it's used would be good,

src/composables/useLeaflet.ts
```diff
@@ -0,0 +1,138 @@
...
+    getTooltip: (f: Feature) => TooltipOptionAndContent,
+    layerEvents: (f: Feature) => LeafletEventHandlerFnMap
+) => {
+    // external refs: map related
+    const map = shallowRef<typeof LMap | null>(null);
+    const getLeafletMap = () => map.value!.leafletObject as Map | undefined;
+
+    // external refs: bounds related
+    const bounds = ref<LatLngBounds | null>(null);
+    const lockBounds = ref<boolean>(false);
```
EmmaLRussell, line 46: I think a comment on what this is and how it will be used would be good.

EmmaLRussell (commented):
Yeah this great, really nice to separate out the Leaflet related stuff so the Choropleth feels more manageable. I like how `useLeaflet` doesn't rely on any other composables - although we have used that pattern elsewhere. 

> added useDataSummary composable since this is completely independent from the functionality, i have not tested this since it is used in the e2e tests, feels weird to test something that is just used for testing idk

I feel a little bit like this should be more tied to the Choropleth, since it was originally meant to be used for testing the choropleth functionality.  However, what that was originally testing has now moved on a bit - i *think* it originally presented a summary of the now removed FeaturesWithColours, a computed property managed by the Choropleth.

It almost feels now that it would actually make a little more sense for the data summary to be derived from what the new `useLeaflet` composable does, in order to make it a useful weather vane of what's actually being shown in the map. Like, maybe it gets build in the course of `updateLeafletMap`. What do you think?


In terms of comments, since useLeaflet is now separated from the Choropleth, I think it would be nice to include a short comment on each of the methods in `useLeaflet` about when it's triggered and why it's doing what it does.

src/utils.ts
M-Kusumgar: i prefer the idea of `choropleth/map/utils` but gone with `composables/utils` for now, typescript complaining about `computed` not being found, etc, not actually sure what the problem is, i skimmed over the tsconfig and vite config files, would rather not mess around with lots of config options :laughing: ticket is here: https://mrc-ide.myjetbrains.com/youtrack/agiles/103-67/current?issue=mrc-5351

src/composables/useLeaflet.ts
M-Kusumgar: done

src/components/Choropleth.vue
M-Kusumgar: done, yh i think this is good because it is kind of like an `@click` event that the normal geojson may expose, it also feels separate because anything to do with the loading or the url of the app should be the choropleths responsibility!

src/composables/useColourScale.ts
M-Kusumgar: yhhh nice, done

david-mears-2:
@EmmaLRussell 
> I like how useLeaflet doesn't rely on any other composables

I'm not seeing the virtue of having `tooltipForFeature` being, effectively, imported by Choropleth and then passed to useLeaflet -- this means that I have to read three files to follow the function to its definition, as opposed to if useLeaflet consumed useTooltip directly. It maybe just creates the _illusion_ of useLeaflet not having dependencies, even though it actually does depend on useTooltip?

Edited to add: Ah, but I see that the call to `tooltipForFeature` uses a couple of utils defined in Choropleth. That's a good reason to call it in that component.

src/composables/useLeaflet.ts
```diff
@@ -0,0 +1,155 @@
...
+
+    // external refs: bounds related
+
+    const bounds = ref<LatLngBounds | null>(null);
+    // toggling lockBounds on before updating the leaflet map
+    // will lock the minZoom and maxBounds (effectively stops the
+    // user from zooming out to the world when they have selected a
+    // country or moving around the map to see different countries).
+
+    // This toggle will be switched the false at the end of every map
```
david-mears-2:
```suggestion
    // This toggle will be switched to false at the end of every map
```

M-Kusumgar:
> I'm not seeing the virtue of having `tooltipForFeature` being, effectively, imported by Choropleth and then passed to useLeaflet -- this means that I have to read three files to follow the function to its definition, as opposed to if useLeaflet consumed useTooltip directly. It maybe just creates the _illusion_ of useLeaflet not having dependencies, even though it actually does depend on useTooltip?

the `useLeaflet` now relies on `useDataSummary` anyway haha but the reason why `getTooltip` exists in choropleth is because it uses the `getFeatureId` and `getFeatureName` from it, having said that, i could extract these to functions in a utils file but then i also wanted `useLeaflet` to be vaguely reminiscent of how you would use leaflet, like an extension of leaflet and how tooltips are displayed is not a leaflet responsibility, it should be the choropleth's responsibility.

so not super sure about what to do

src/composables/useLeaflet.ts
```diff
@@ -0,0 +1,169 @@
...
+    // are added to the map to set the view and zoom of the map
+    const updateRegionBounds = (regionId?: string) => {
+        const leafletMap = getLeafletMap();
+        const countryBounds = getRegionBounds(regionId);
+
+        if (leafletMap && countryBounds) {
+            bounds.value = countryBounds;
+            leafletMap.fitBounds(bounds.value);
+        }
+    };
```
david-mears-2:
```suggestion
    const updateRegionBounds = (regionId?: string) => {
        const leafletMap = getLeafletMap();
        const regionBounds = getRegionBounds(regionId);

        if (leafletMap && regionBounds) {
            bounds.value = regionBounds;
            leafletMap.fitBounds(bounds.value);
        }
    };
```

src/composables/useLeaflet.ts
```diff
@@ -0,0 +1,169 @@
...
+    GeoJSONOptions
+} from "leaflet";
+import { storeToRefs } from "pinia";
+import { useAppStore } from "../stores/appStore";
+import { minZoom } from "./utils";
+import { useDataSummary } from "./useDataSummary";
+
+type FeatureProperties = { GID_0: string; GID_1: string; NAME_1: string };
+type TooltipOptionAndContent = { content: string; options?: TooltipOptions };
+// laeflet geojson options do not include smoothFactor, probably outdated
```
david-mears-2:
```suggestion
// leaflet geojson options do not include smoothFactor, probably outdated
```

david-mears-2 (approved):
Thanks for adding all the inline comments!

src/composables/useColourScale.ts
```diff
@@ -11,6 +12,11 @@ interface IndicatorRange {
     range: number;
 }
 
+type FillAndOutlineColour = {
+    fillColor: string;
+    color: string;
```
david-mears-2: It's not clear, if I'm somewhere else in the codebase that's using useColourScale, what the difference between 'fillColor' and 'color' is meant to be -- unless I read this file carefully.

M-Kusumgar: can add a comment there! i only chose them so i can spread into the style object haha

M-Kusumgar: done

src/composables/useColourScale.ts
```diff
@@ -13,7 +13,9 @@ interface IndicatorRange {
 }
 
 type FillAndOutlineColour = {
+    // the fill colour of the feature
     fillColor: string;
+    // the outline colour of the feature
     color: string;
 };
```
david-mears-2: My preference would be to rename `color` to `outlineColor`, as opposed to making people hunt for the inline comment. Sorry it's a bit tedious to rename.

src/utils.ts
EmmaLRussell: Sorry, I completely mistyped there, I meant `components/utils` or `components/map/utils`! :facepalm:
Shouldn't be in composables!

EmmaLRussell:
> @david-mears-2 
> I'm not seeing the virtue of having `tooltipForFeature` being, effectively, imported by Choropleth and then passed to useLeaflet -- this means that I have to read three files to follow the function to its definition, as opposed to if useLeaflet consumed useTooltip directly. It maybe just creates the _illusion_ of useLeaflet not having dependencies, even though it actually does depend on useTooltip?

I think what I like about it is that the composables don't directly depend on eachother - they do have a dependency but they're not assuming what it is, and it's the Choropleth, as the client of all the composables, which coordinates them together, effectively doing dependency injection. Theoretically easier to test.. if we were testing them! :laughing:

EmmaLRussell (approved):
Thanks, that's a lot clearer with the comments!

src/utils.ts
M-Kusumgar: i was thinking it was an odd choice but with ides im not so fussed about folder structure so i was like sure :laughing:, have moved it to components/utils

src/composables/useColourScale.ts
M-Kusumgar: no worries! done

PR #22, opened 2024-05-08
Mrc 5200 Add admin level toggle

M-Kusumgar:
This adds the admin level toggle to the map, which will only appear when you have selected a country.

Main changes:
* new `routerPush` utility with consistent debounce timeout
* new `AdminLevelToggle` component

Need to still do tests

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -30,7 +30,8 @@ export default defineConfig({
         /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
         trace: "on-first-retry",
 
-        headless: true
+        headless: true,
+        screenshot: "only-on-failure"
     },
 
     /* Configure projects for major browsers */
--- a/src/components/AdminLevelToggle.vue
+++ b/src/components/AdminLevelToggle.vue
@@ -1,0 +1,22 @@
+<template>
+    <v-btn-toggle id="admin-toggle" :model-value="mapSettings.adminLevel" @update:model-value="emit('change-admin-level', $event)" mandatory :max="1" :density="'compact'" theme="dark">
+        <v-btn :value="1" :ripple="false" selected-class="selected-button">Admin 1</v-btn>
+        <v-btn :value="2" :ripple="false" :disabled="admin2DataMissing" selected-class="selected-button">Admin 2</v-btn>
+    </v-btn-toggle>
+</template>
+
+<script setup lang="ts">
+import { storeToRefs } from "pinia";
+import { useAppStore } from "../stores/appStore";
+
+const { mapSettings, appConfig } = storeToRefs(useAppStore());
+
+const admin2DataMissing = computed(() => appConfig.value.countriesWithoutAdmin2.includes(mapSettings.value.country));
+
+const emit = defineEmits(["change-admin-level"]);
+</script>
+<style scoped>
+.selected-button {
+    font-weight: bolder;
+}
+</style>
--- a/src/components/AdminLevelToggle.vue
+++ b/src/components/AdminLevelToggle.vue
@@ -1,5 +1,13 @@
 <template>
-    <v-btn-toggle id="admin-toggle" :model-value="mapSettings.adminLevel" @update:model-value="emit('change-admin-level', $event)" mandatory :max="1" :density="'compact'" theme="dark">
+    <v-btn-toggle
+        id="admin-toggle"
+        :model-value="mapSettings.adminLevel"
+        @update:model-value="emit('change-admin-level', $event)"
+        mandatory
+        :max="1"
+        :density="'compact'"
+        theme="dark"
+    >
         <v-btn :value="1" :ripple="false" selected-class="selected-button">Admin 1</v-btn>
         <v-btn :value="2" :ripple="false" :disabled="admin2DataMissing" selected-class="selected-button">Admin 2</v-btn>
     </v-btn-toggle>
--- a/src/components/Choropleth.vue
+++ b/src/components/Choropleth.vue
@@ -15,6 +15,9 @@
             <LControl position="topleft">
                 <ResetMapButton :selected-indicator="mapSettings.indicator" @reset-view="updateRegionBounds" />
             </LControl>
+            <LControl position="topright" v-if="mapSettings.country">
+                <AdminLevelToggle @change-admin-level="handleChangeAdminLevel" />
+            </LControl>
         </LMap>
         <div style="visibility: hidden" class="choropleth-data-summary" v-bind="dataSummary"></div>
     </div>
@@ -31,10 +34,11 @@ import { useLeaflet } from "../composables/useLeaflet";
 import "leaflet/dist/leaflet.css";
 import { useTooltips } from "../composables/useTooltips";
 import { APP_BASE_ROUTE } from "../router/utils";
-import { debounce } from "../utils";
+import { routerPush, AdminLevel } from "../utils";
 import { backgroundLayer } from "./utils";
 import { useLoadingSpinner } from "../composables/useLoadingSpinner";
 import { useSelectedMapInfo } from "../composables/useSelectedMapInfo";
+import AdminLevelToggle from "./AdminLevelToggle.vue";
 
 const mapLoading = ref(true);
 const router = useRouter();
@@ -61,6 +65,13 @@ const getFeatureName = (feature: Feature) =>
         ? feature.properties[featureProperties.nameAdm2]
         : feature.properties[featureProperties.nameAdm1];
 
+const handleChangeAdminLevel = (level: number) => {
+    mapLoading.value = true;
+    const { indicator, country } = mapSettings.value;
+    const adminLevel = level === 1 ? AdminLevel.ONE : AdminLevel.TWO;
+    routerPush(router, `/${APP_BASE_ROUTE}/${indicator}/${country}/${adminLevel}`);
+};
+
 const style = (f: Feature) => {
     const { country, indicator } = mapSettings.value;
     const isFaded = !!country && !featureInSelectedCountry(f);
@@ -80,7 +91,7 @@ const layerOnEvents = (feature: Feature) => {
             const country = feature.properties[featureProperties.country];
             // select feature's country, or unselect if click on it when already selected
             const countryToSelect = country === mapSettings.value.country ? "" : country;
-            debounce(() => router.push(`/${APP_BASE_ROUTE}/${mapSettings.value.indicator}/${countryToSelect}`))();
+            routerPush(router, `/${APP_BASE_ROUTE}/${mapSettings.value.indicator}/${countryToSelect}`);
         }
     };
 };
@@ -93,9 +104,7 @@ const { map, dataSummary, lockBounds, updateLeafletMap, handleMapBoundsUpdated, 
 useLoadingSpinner(map, mapLoading);
 
 const updateMap = () => {
-    if (mapSettings.value.country) {
-        lockBounds.value = true;
-    }
+    lockBounds.value = !!mapSettings.value.country;
     updateLeafletMap(selectedFeatures.value, mapSettings.value.country);
 };
 
--- a/src/pages/index.vue
+++ b/src/pages/index.vue
@@ -101,9 +101,12 @@ const selectDataForRoute = async () => {
     if (country) {
         const admin2DataMissing = appConfig.value.countriesWithoutAdmin2.includes(country);
         if (!adminLevel) {
-            router.replace(`/${pathogen}/${version}/${indicator}/${country}/${admin2DataMissing ? AdminLevel.ONE : AdminLevel.TWO}`);
+            router.replace(
+                `/${pathogen}/${version}/${indicator}/${country}/${admin2DataMissing ? AdminLevel.ONE : AdminLevel.TWO}`
+            );
             return;
-        } else if (adminLevel === AdminLevel.TWO && admin2DataMissing) {
+        }
+        if (adminLevel === AdminLevel.TWO && admin2DataMissing) {
             router.replace(`/${pathogen}/${version}/${indicator}/${country}/${AdminLevel.ONE}`);
             return;
         }
--- a/src/pages/index.vue
+++ b/src/pages/index.vue
@@ -17,7 +17,7 @@ import { useRouter } from "vue-router";
 import { useAppStore } from "../stores/appStore";
 import NotFound from "./notFound.vue";
 import { APP_BASE_ROUTE, PATHOGEN, VERSION } from "../router/utils";
-import { mapSettingsAreEqual } from "../utils";
+import { mapSettingsAreEqual, stringAdminLevelToNumeric, AdminLevel } from "../utils";
 import { MapSettings } from "../types/resourceTypes";
 
 const router = useRouter();
@@ -41,6 +41,11 @@ const props = defineProps({
         default: ""
     },
     country: {
+        type: String,
+        required: false,
+        default: ""
+    },
+    adminLevel: {
         type: String,
         required: false,
         default: ""
@@ -71,7 +76,8 @@ const selectDataForRoute = async () => {
         pathogen: [PATHOGEN],
         version: [VERSION],
         indicator: Object.keys(appConfig.value.indicators),
-        country: appConfig.value.countries
+        country: appConfig.value.countries,
+        adminLevel: Object.keys(stringAdminLevelToNumeric)
     };
     const propsWithCorrectCase: Partial<Record<PropName, string>> = {};
     Object.keys(possibleValuesForProps).forEach((prop: PropName) => {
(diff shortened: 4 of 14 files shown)


src/utils.ts
```diff
@@ -10,6 +12,10 @@ export function debounce(fn, wait = 0) {
     };
 }
 
+export const routerPush = (router: Router, path: string) => {
+    debounce(() => router.push(path), 75)();
```
M-Kusumgar, line 16: 75 ms seems to work well for showing loading spinner, having no timeout wasn't working well with admin level toggle so thought to extract this out so that it is consistent with all the `router.push`'s

david-mears-2:
The spinner turns on when you click 'admin2' if you're already on an admin2 view, giving the appearance of hanging.

src/components/Choropleth.vue
```diff
@@ -48,15 +52,20 @@ const featureInSelectedCountry = (feature: Feature, selectedCountry: string) =>
     feature.properties[featureProperties.country] === selectedCountry;
 
 const getFeatureId = (feature: Feature) =>
-    featureInSelectedCountry(feature, selectedCountryId.value)
+    featureInSelectedCountry(feature, selectedCountryId.value) && selectedAdminLevel.value !== 1
```
david-mears-2: I'm not sure I understand this change. Was it not fully working before?

src/pages/index.vue
```diff
@@ -3,7 +3,7 @@
         <Choropleth v-if="selectedIndicator" data-testid="choropleth" />
         <div class="sticky-footer">
             <div v-for="name in indicatorNames" :key="name">
-                <router-link :to="`/${APP_BASE_ROUTE}/${name}/${selectedCountryId}`" custom v-slot="{ navigate }">
+                <router-link :to="getRouterLink(name)" custom v-slot="{ navigate }">
```
david-mears-2: :+1:

src/components/Choropleth.vue
M-Kusumgar: this would give you the admin level 2 feature ids when admin level 1 is selected, but we want admin 1 ids when it is selected otherwise the tooltips would be wrong

david-mears-2 (approved):

src/components/AdminLevelToggle.vue
```diff
@@ -0,0 +1,34 @@
+<template>
+    <div id="admin-toggle">
+        <v-btn
+            @click="() => handleClick(1)"
+            role="link"
+            class="mb-2 rounded-s-lg rounded-e-0 toggle-btn"
+            :class="selectedAdminLevel === 1 ? 'bg-blue' : 'bg-black'"
```
EmmaLRussell: Could use primary colour from the theme here, via "v-theme..." css variable

tests/unit/components/adminLevelToggle.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+        const { findAllByRole } = renderComponent();
+        const buttons = await findAllByRole("link");
+        expect(buttons.length).toBe(2);
+        const [button1, button2] = buttons;
+        expect(button1).toHaveClass("mb-2 rounded-s-lg rounded-e-0 toggle-btn bg-blue");
+        expect(button1).toHaveTextContent("Admin 1");
+        expect(button2).toHaveClass("mb-2 rounded-s-0 rounded-e-lg toggle-btn bg-black");
+        expect(button2).toHaveTextContent("Admin 2");
+    });
+});
```
EmmaLRussell, line 52: Could add a test for emit on click..

src/components/Choropleth.vue
```diff
@@ -75,7 +84,10 @@ const layerOnEvents = (feature: Feature) => {
             const country = feature.properties[featureProperties.country];
             // select feature's country, or unselect if click on it when already selected
             const countryToSelect = country === selectedCountryId.value ? "" : country;
-            debounce(() => router.push(`/${APP_BASE_ROUTE}/${selectedIndicator.value}/${countryToSelect}`))();
+            const path = countryToSelect
+                ? `/${APP_BASE_ROUTE}/${selectedIndicator.value}/${countryToSelect}/admin2`
```
EmmaLRussell: This should be OK to keep as country route without admin level defined in the path - it should default to admin2

src/components/Choropleth.vue
```diff
@@ -99,5 +109,11 @@ const finishUpdatingMap = () => {
     mapLoading.value = false;
 };
 
-watch([selectedFeatures, selectedIndicator], ([newFeatures]) => updateMap(newFeatures));
+watch(
+    [selectedFeatures, selectedIndicator, selectedAdminLevel],
+    ([newFeatures, , newAdminLevel], [, , oldAdminLevel]) => {
+        const doLockBounds = !!selectedCountryId.value || newAdminLevel !== oldAdminLevel;
```
EmmaLRussell: why do we need to change the bounds lock if the admin level has changed - would it remove the previous lock otherwise?

src/pages/index.vue
```diff
@@ -51,9 +53,20 @@ const props = defineProps({
...
+        required: false,
+        default: ""
     }
 });
 
+const getRouterLink = (name: string) => {
+    return selectedCountryId.value
+        ? `/${APP_BASE_ROUTE}/${name}/${selectedCountryId.value}/${props.adminLevel}`
+        : `/${APP_BASE_ROUTE}/${name}/`;
+};
```
EmmaLRussell: As I say, I think it would be better to default to cleaner url without admin2 part (which can optionally by provided) if there's a selected country and admin level is 2.

src/pages/index.vue
```diff
@@ -51,9 +53,20 @@ const props = defineProps({
...
         default: ""
+    },
+    adminLevel: {
+        type: String,
+        required: false,
+        default: ""
     }
 });
 
+const getRouterLink = (name: string) => {
```
EmmaLRussell:
```suggestion
const getRouterLink = (indicatorId: string) => {
```

EmmaLRussell: This could potentially be a more flexible util (or maybe a couple of utils_ which could be used by other  components which build routes e.g. Choropleth - would need to include options for either setting a new selected country or keeping the existing one though...

src/utils.ts
```diff
@@ -9,3 +11,14 @@ export function debounce(fn, wait = 0) {
         }, wait);
     };
 }
+
+export const routerDebounce = 75;
```
EmmaLRussell, line 32: Why 75? Does it not work just with min debounce?

EmmaLRussell (commented):
Nice! Bit of a styling clash with the indicator menu branch, which is indicating selection by secondary colour border rather than primary colour background. I think background does make more sense for the toggle though, so maybe I'll just make the menu use the primary colour from the theme. 

We should be able to browse to country url without admin level defined eg dengue/may24/FOI/BRA. In this case, admin2 regions are correctly displayed, but the toggle does not show Admin 2 as selected:
image

There's a lot of codecov grumbles - I know we're being fairly relaxed about Choro testing, but I think there's probably more we could in other modules.

david-mears-2:
> Bit of a styling clash with the indicator menu branch, which is indicating selection by secondary colour border rather than primary colour background. I think background does make more sense for the toggle though, so maybe I'll just make the menu use the primary colour from the theme.

I actually find the toggle background colours unclear - it isn't obvious to me which one out of blue or black means 'selected' vs 'unselected'. If it had a colour border, then I could read more easily which one is being highlighted. (Optionally also we should make the one that is selected be not a link, otherwise it's inviting you to click on something that makes no change -- that would also help, and it would work if we decided not to use borders.)

EmmaLRussell:
> > Bit of a styling clash with the indicator menu branch, which is indicating selection by secondary colour border rather than primary colour background. I think background does make more sense for the toggle though, so maybe I'll just make the menu use the primary colour from the theme.
> 
> I actually find the toggle background colours unclear - it isn't obvious to me which one out of blue or black means 'selected' vs 'unselected'. If it had a colour border, then I could read more easily which one is being highlighted. (Optionally also we should make the one that is selected be not a link, otherwise it's inviting you to click on something that makes no change -- that would also help, and it would work if we decided not to use borders.)

I'd be on board with using a toggle here too, it would be more consistent...

tests/unit/components/adminLevelToggle.spec.ts
```diff
@@ -0,0 +1,29 @@
+import { render } from "@testing-library/vue";
+import { describe, expect, test } from "vitest";
+import { createVuetify } from "vuetify";
+import { mockPinia } from "../mocks/mockPinia";
+import AdminLevelToggle from "../../../src/components/AdminLevelToggle.vue";
+
+const store = mockPinia({ selectedAdminLevel: 1 });
+const renderComponent = () => {
+    const vuetify = createVuetify();
```
david-mears-2: I think this is duplicating `mockVuetify`.

src/components/Choropleth.vue
M-Kusumgar: yep i have gone with resetting the lock every update and then explicitly defining when we want to lock, think it is better if anything goes wrong that the bounds are the whole world

tests/unit/components/adminLevelToggle.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+    });
+};
+
+describe("AdminLevelToggle", () => {
+    test("renders as expected", async () => {
+        const { findAllByRole } = renderComponent();
+        const buttons = await findAllByRole("link");
+        expect(buttons.length).toBe(2);
+        const [button1, button2] = buttons;
+        expect(button1).toHaveClass("mb-2 rounded-s-lg rounded-e-0 toggle-btn bg-blue");
```
david-mears-2: I'm not used to such specific tests as this. What's the value in testing for classes?

src/components/AdminLevelToggle.vue
M-Kusumgar: using the theme now

tests/unit/components/adminLevelToggle.spec.ts
M-Kusumgar: done

src/components/Choropleth.vue
M-Kusumgar: yep done

src/pages/index.vue
M-Kusumgar: for now left this because we no longer need this function since the default is to add admin2 to the route if not specified

src/utils.ts
M-Kusumgar: yh it just seemed to be consistent on my machine

tests/unit/components/adminLevelToggle.spec.ts
M-Kusumgar: yep this was too specific, have removed it

M-Kusumgar:
> > Bit of a styling clash with the indicator menu branch, which is indicating selection by secondary colour border rather than primary colour background. I think background does make more sense for the toggle though, so maybe I'll just make the menu use the primary colour from the theme.
> 
> I actually find the toggle background colours unclear - it isn't obvious to me which one out of blue or black means 'selected' vs 'unselected'. If it had a colour border, then I could read more easily which one is being highlighted. (Optionally also we should make the one that is selected be not a link, otherwise it's inviting you to click on something that makes no change -- that would also help, and it would work if we decided not to use borders.)

So i tried adding a border and that was not working because the toggle component is just two buttons and they have some styling from the theme that cuts off the edge they share, so the border is only on the outside edges of the button in the toggle, i have however made the selected button text bold, let me know if that is clearer!

david-mears-2:
Looks really slick with the monochrome styles!

Maybe for a separate ticket: when you are on admin 1 level looking at country A, and you click to see country B, the admin level switches to 2 (because that is our default). Is that what we want / what the user would expect?

Relatedly, and less importantly, when you alternate between 'reset map' and clicking on a country, again the admin level always prefers level 2 even if you've just recently been using level 1; this makes sense if level 2 is the strong default, but another way of looking at it would be to say persist the admin level and switch it only when the user asks it to.

src/utils.ts
```diff
@@ -17,3 +18,19 @@ export const mapSettingsAreEqual = (mapSettings1: MapSettings, mapSettings2: Map
...
+
+export enum AdminLevel {
+    ONE = "admin1",
+    TWO = "admin2"
+}
+
+export const stringAdminLevelToNumeric = {
+    [AdminLevel.ONE]: 1,
+    [AdminLevel.TWO]: 2
+};
```
david-mears-2, line 30: What do square brackets do here?

tests/e2e/index.spec.ts
```diff
@@ -162,4 +162,20 @@ test.describe("Index page", () => {
         await page.waitForURL(/dengue\/may24\/serop9/);
         await expect(await page.locator(".leaflet-control-zoom-out.leaflet-disabled")).toHaveCount(1);
     });
+
+    test("admin toggle works", async ({ page }) => {
```
david-mears-2:
```suggestion
    test("admin level toggle works", async ({ page }) => {
```

tests/e2e/index.spec.ts
```diff
@@ -162,4 +162,20 @@ test.describe("Index page", () => {
...
+        const allRegions = await page.locator(GEOJSON_SELECTOR);
```
david-mears-2, line 171: Maybe include an expectation that it's not displayed on the home view

tests/unit/components/adminLevelToggle.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+        const [, admin2Button] = buttons;
+        expect(admin2Button).toHaveClass("selected-button");
+    });
+
+    test("admin level 2 is disabled if country missing admin 2 data", async () => {
+        const { findAllByRole } = renderComponent(2, true);
+        const buttons = await findAllByRole("button");
+        const [, admin2Button] = buttons;
+        expect(admin2Button).toBeDisabled();
+    });
```
david-mears-2:
```suggestion
    test("admin level 2 is disabled if country missing admin 2 data", async () => {
        const { findAllByRole } = renderComponent(2, true);
        const buttons = await findAllByRole("button");
        const [admin1Button, admin2Button] = buttons;
        expect(admin1Button).toHaveClass("selected-button");
        expect(admin2Button).toBeDisabled();
    });
```

david-mears-2: ^ adding a check that admin1 is selected in this scenario

src/pages/index.vue
```diff
@@ -86,18 +92,32 @@ const selectDataForRoute = async () => {
...
         return;
     }
+    if (country) {
+        const admin2DataMissing = appConfig.value.countriesWithoutAdmin2.includes(country);
+        if (!adminLevel) {
+            router.replace(
+                `/${pathogen}/${version}/${indicator}/${country}/${admin2DataMissing ? AdminLevel.ONE : AdminLevel.TWO}`
+            );
+            return;
+        }
```
david-mears-2, line 108: Is this worth an e2e test? (genuine question)

src/stores/appStore.ts
```diff
@@ -48,14 +48,14 @@ export const useAppStore = defineStore("app", {
...
-                state.admin2Indicators[country] = state.admin1Indicators[country];
-            }
-
-            if (level === 2 && !admin2DataMissing) {
-                state.admin2Indicators[country] =
-                    state.admin2Indicators[country] || (await getIndicators(country, level));
+            if (level === 2) {
+                if (admin2DataMissing) {
+                    level = 1;
+                    state.admin2Indicators[country] = state.admin1Indicators[country];
```
david-mears-2, line 54: We should probably have a unit test on this.

src/utils.ts
M-Kusumgar: these let you use variables as keys, so it will have key `"admin1"` and value `1` in the first row, it's nice because we can just defined the "admin1" string in one place

tests/e2e/index.spec.ts
M-Kusumgar: done

src/pages/index.vue
M-Kusumgar: yh probably worth test if a country is missing data then admin1 is in url, i dont think it is worth explicitly testing the flip side though

src/stores/appStore.ts
M-Kusumgar: yep will probably opt for doing this opportunistically in another PR, i havent changed any logic here, just changed style of code

EmmaLRussell:
> Looks really slick with the monochrome styles!
> 
> Maybe for a separate ticket: when you are on admin 1 level looking at country A, and you click to see country B, the admin level switches to 2 (because that is our default). Is that what we want / what the user would expect?
> 
> Relatedly, and less importantly, when you alternate between 'reset map' and clicking on a country, again the admin level always prefers level 2 even if you've just recently been using level 1; this makes sense if level 2 is the strong default, but another way of looking at it would be to say persist the admin level and switch it only when the user asks it to.

We've never defined what this behaviour should be I think either would be acceptable, would probably favour the current behaviour of defaulting to admin2 when you go to a fresh country.

playwright.config.ts
```diff
@@ -30,7 +30,8 @@ export default defineConfig({
         /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
         trace: "on-first-retry",
 
-        headless: true
+        headless: true,
+        screenshot: "only-on-failure"
```
EmmaLRussell, line 34: I take it the e2e tests were a journey then? 😆

EmmaLRussell (approved):
Works great! The selected styling on the toggle is quite subtle, but looks good. 

I see you've disabled the Admin2 button when there's no admin 2 data available (rather than removing the toggle). Might be worth a tooltip to explain that when it's disabled?

playwright.config.ts
M-Kusumgar: ahh i was just being silly really but also is a nice to have

M-Kusumgar:
> I see you've disabled the Admin2 button when there's no admin 2 data available (rather than removing the toggle). Might be worth a tooltip to explain that when it's disabled?

i tried to just get it in for this PR thinking it would be simple but turns out to be a little fiddly because vuetify tooltips dont appear on disabled buttons, im sure its still not super difficult but have split it into another ticket considering it wont just be a one line change: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-5397/Adding-tooltip-to-admin-2-disabled-button

PR #24, opened 2024-05-09
Make leaflet attribution responsive

david-mears-2:
Just a little thing I noticed on small screens. With iPhone SE, the leaflet attribution takes up two lines of real estate. Here we change the font-size so it's only one line on smaller screens.

Screenshot from 2024-05-09 14-47-11
Screenshot from 2024-05-09 14-52-09

### DIFF
--- a/src/styles/settings.scss
+++ b/src/styles/settings.scss
@@ -49,3 +49,8 @@
   }
 }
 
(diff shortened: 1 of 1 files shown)


EmmaLRussell (approved):
nice!

PR #25, opened 2024-05-09
Mrc 5352 Fix tooltips not appearing

M-Kusumgar:
if you go away from a region and hover back on to the same region without hovering over any other regions (easiest to do in for regions that have a coast) then the tooltip does not appear. this pr fixes that

### DIFF
--- a/src/composables/useLeaflet.ts
+++ b/src/composables/useLeaflet.ts
@@ -118,8 +118,10 @@ export const useLeaflet = (
                 // map while the bounds were locked, now we track each layer that will open a
                 // tooltip and close them if they are not the most recent layer to make sure
                 // no old tooltips remain open
-                layerWithOpenTooltip.value?.closeTooltip();
-                layerWithOpenTooltip.value = layer;
+                if (layer !== layerWithOpenTooltip.value) {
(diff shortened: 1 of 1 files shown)


david-mears-2 (approved):
It works!!

src/composables/useLeaflet.ts
```diff
@@ -92,7 +92,9 @@ export const useLeaflet = (
                 // map while the bounds were locked, now we track each layer that will open a
                 // tooltip and close them if they are not the most recent layer to make sure
                 // no old tooltips remain open
-                layerWithOpenTooltip.value?.closeTooltip();
+                if (layer !== layerWithOpenTooltip.value) {
+                    layerWithOpenTooltip.value?.closeTooltip();
+                }
```
EmmaLRussell: Presumably you could include line 98 in the if clause too - no point in setting layerWithOpenTooltip value to layer if you've already check that it is...

EmmaLRussell (approved):
works for me...

src/composables/useLeaflet.ts
M-Kusumgar: yep done

PR #26, opened 2024-05-09
mrc-5353 Reverse Force of Infection colour scale

EmmaLRussell:
Reverse scale so lower FOI corresponds to cooler colours - as requested by Clare


david-mears-2 (approved):
Alles gut!

PR #27, opened 2024-05-10
mrc-5214 User instructions

david-mears-2:
These instructions should be visible at first, and then if the user ever dismisses them, they have to click the right button to get the help text back. This preference will be persisted between sessions in localStorage. (You can evade that by using a private browsing window such as incognito, allowing you to test it.)

I wonder if this merits an e2e test on top of the component test? To verify that the alert is present on the first visit, and stays dismissed once dismissed when you navigate to a new page.

Rebased onto the admin level toggle branch because this is using the same top-right corner of the map as that branch.

### DIFF
--- a/tests/unit/components/helpAlert.spec.ts
+++ b/tests/unit/components/helpAlert.spec.ts
@@ -9,31 +9,47 @@ const renderComponent = () => {
     });
 };
 
+let helpAlertHasBeenDismissed: string | null;
+const localStorageMock = (() => {
+    return {
+        getItem: vi.fn(() => helpAlertHasBeenDismissed),
+        setItem: vi.fn()
+    };
+})();
+
+const originalLocalStorage = (window as any).localStorage;
+beforeAll((): void => {
+    (window as any).localStorage = localStorageMock;
+});
+afterAll((): void => {
+    (window as any).localStorage = originalLocalStorage;
+});
+
 describe("HelpAlert", () => {
     describe("When the alert has not been dismissed before", () => {
+        beforeAll((): void => {
+            helpAlertHasBeenDismissed = null;
+        });
+
         it("should show the alert", () => {
             renderComponent();
             expect(screen.getByRole("alert")).toBeVisible();
         });
+
+        describe("When the alert is dismissed", () => {
+            it("should call localStorage.setItem and not show alert", async () => {
+                renderComponent();
+                const closeButton = screen.getByLabelText("Close");
+                await fireEvent.click(closeButton);
+                expect(localStorageMock.setItem).toHaveBeenCalled();
+                expect(screen.queryByRole("alert")).toBeNull();
+            });
+        });
     });
 
     describe("When the alert was dismissed in the past", () => {
-        let originalLocalStorage;
-
         beforeAll((): void => {
-            const localStorageMock = (() => {
-                return {
-                    getItem: vi.fn(() => "true"),
-                    setItem: vi.fn()
-                };
-            })();
-
-            originalLocalStorage = window.localStorage;
-            (window as any).localStorage = localStorageMock;
-        });
-
-        afterAll((): void => {
-            (window as any).localStorage = originalLocalStorage;
+            helpAlertHasBeenDismissed = "true";
         });
 
         it("should not show the alert", () => {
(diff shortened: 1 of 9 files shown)


tests/unit/components/helpAlert.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+    });
+
+    describe("When the alert was dismissed in the past", () => {
+        let originalLocalStorage;
+
+        beforeAll((): void => {
+            const localStorageMock = (() => {
+                return {
+                    getItem: vi.fn(() => "true"),
+                    setItem: vi.fn()
```
EmmaLRussell: since you're setting this up anyway, could also check that it calls setItem when the alert is dismissed...

EmmaLRussell (commented):
Looks great, works perfectly. 

Agree that an e2e test would be good. 

I don't think the help button should go below the admin level toggle, which doesn't always appear -feels like it should be above, or to the right.

I also think it might be confusing that the admin level described in the text isn't visible when the text can be read... not sure what's the best way to deal with that though - possibly to put the help text below the toggle...? (Which would imply that you'd also be able to see the help button while help text is visible...)

The text doesn't extend to the right edge of the box - this is particularly noticeable on mobile. Might be worth tweaking to use that space..
image

Some failing tests...

david-mears-2:
After some extensive re-styling where I was trying to avoid using screen space inefficiently, I've pushed a commit that has responsive styling (the alert is shunted into a different gap depending on screen size). This uses the flex attribute 'order' to achieve its goals, and as a result you'll see the `?` **button** that opens the alert sits to the _left_ of the admin toggle on large screens in order to have the **alert** do so when it's opened. I hope that's acceptable - the issue, if any, would be that it means the `?` button changes position when the admin toggle is added/removed on navigating to/from country-level/global view. The reason I've landed on that suboptimal situation is because the alternative is to have the admin-toggle shunt back and forth as you open and close the alert (by putting `order: 3` on `#helpButton`), as seen below ('alternative option'):

### Alternative option

[Screencast from 14-05-24 15:57:40.webm](https://github.com/mrc-ide/arbomap/assets/60350599/21dc17f7-ad39-41db-ab42-b71b670a5919)

### Status quo as of commit '[Use space more efficiently with responsive styles](https://github.com/mrc-ide/arbomap/pull/27/commits/91713b644296e4da3a7e1637901de67ff0a5a2fa)'

[Screencast from 14-05-24 15:59:09.webm](https://github.com/mrc-ide/arbomap/assets/60350599/baef19b1-ea1d-43b1-8f9c-9be99a48700a)

david-mears-2:
Current test failures are inherited from parent [mrc-5200 branch](https://github.com/mrc-ide/arbomap/pull/22)

EmmaLRussell (commented):
Definitely looks better! 

>  the alternative is to have the admin-toggle shunt back and forth as you open and close the alert (by putting `order: 3` on `#helpButton`)

Yeah, the shunting isn't great, but neither is the Help button hopping around. I do feel that the Help button should always be closer to the top/right than the admin toggle, as it's not a direct controller of the app so should be more peripheral. 

Another alternative would be to let the Help button remain visible when the Help text is being shown (I think that would be fine), or to hide it but not move the toggle. The downside of *that* is that you've then got some unused space (bad on small mobiles), but at least everything stays put. 

OR help button could be an icon in the nav bar  - then nothing would need to hop around and not space would be lost - but it may be a bit of an odd location..?

EmmaLRussell:
> on pause, converted to draft, will continue to rebase

Sorry! Let's see what @M-Kusumgar thinks, and agree on what best approach would be. Could potentially merge as is to get something in and working and tweak it later.

david-mears-2:
I am cutting my losses with the eternal styling dilemmas as we have discussed moving the admin toggle anyway! So I hope you can live with this for now. As of latest commit "Fix and simplify styles after rebasing", the help alert button is positioned second, after the admin toggle, which has some downsides (cf. Emma's comment about 'peripheral') but means that it doesn't make the admin toggle jump about.

EmmaLRussell (approved):
Looks good! As discussed, we'll revamp the selection controls in a separate branch.

M-Kusumgar (approved):
works for me, happy to merge this in

PR #28, opened 2024-05-10
mrc-5343-latest-data Data update: fill in missing admin1, and include all IND regions

EmmaLRussell:
This branch contains updated dataset, which:
- fills in missing admin1 regions e.g. should not see any missing values in Puerto Rico (PRI)
- includes territories assigned to Z0.. ISOs in previous data set e.g. Kashmir's sub-regions - these are now included in IND, since GADM assigns these regions to India
- adds spaces to all regions with multiple words (where spaces were removed in default GADM Geojson - e.g. Baja California 
- attempts to make sense of Uruguay (URY) - where some areas don't have an admin2 region, but some do. The ones that don't should now say "(no municipality)" and display admin1 values e.g. Artigas in NW of Uruguary. 

We also deliberately exclude certain countries which do have values in the dataset, but where we know those values are derived from regions in neighbouring countries (in `scripts/process-data.R`).

I've also added a script to generate a mapping of country IDs to country names - this will be useful for both Excel download and the country selector.  And I have also added a note to the README attempting to describe the various data files and processing scripts. I've left it a bit mysterious as to where the simplified geojson files come from. Wes has promised to add his scripts to a github repo at some point - when this happens we can add a link here.


README.md
```diff
@@ -17,4 +17,40 @@ Lint: `npm run lint` or `npm run lint-fix`
...
+using the file `data/raw/all_adm1_0_5pc.json`, which contains all features in the form of a single FeatureCollection. We load this
+when the app first loads, to display the global view with admin1 data from all countries.
+- `geojson/admin0`: country-level geojson, one file per country. We use these to draw an outline around the selected country,
+- `geojson/admin1`: admin1-level geojson, one file per country. We keep this only for the small set of countries which have no admin2 regions (`countriesWithoutAdmin2`
+in config.json). This geojson is more detailed than the simplified regions defined in the `all_adm1..` file. When one of these
+countries is selected we load and display from these files.
+- `geojson/admin2`: admin2-level geojson, one file per country. When a country is selected, we load and display from these files. 
+- `indicators/admin1`: a single file containing all admin1 level indicator values. We load this when the app first loads, to display
+the global view. Generated from the xlsx file in `data/raw` by the script `process-data.R`
+- `indicators/admin2`: admin2 level indicator values, one file per country, loaded when the country is selected. Generated
```
M-Kusumgar:
```suggestion
- `indicators/admin2`: admin2-level indicator values, one file per country, loaded when the country is selected. Generated 
```

scripts/generate_country_names_from_adm0_geojson.cjs
```diff
@@ -0,0 +1,41 @@
+/* eslint-disable import/no-dynamic-require, global-require */
+
+// We need country names for Excel download and Country selector, but these are not reliably present in the processed
+// indicator files or simplified admin1 or admin2 geojson. They are present in the admin0 geojson, but we don't want to
+// have to read in all these files in-app order to find all country names. So we generate pre-canned names from ammin0
```
M-Kusumgar:
```suggestion
// have to read in all these files in order to find all country names. So we generate pre-canned names from admin0
```

M-Kusumgar (approved):
looks good to me, just a couple of comment changes but works well!

PR #30, opened 2024-05-14
mrc-5170 Excel download

EmmaLRussell:
This branch adds Excel download functionality to the app. There's a new button in the top right of the choropleth - this should look less odd when the admin level toggle is moved to the main menu. 

The Excel file which is downloaded will include indicator values for:
- all admin1 features globally, if no country is selected (in admin1 tab)
- admin1 and admin2 features for the selected country, if any (in admin1 and admin2 tabs - or just admin1 if it's one of those countries which only has admin1 regions)

The columns in the file consist of
- ids and names for all country and subcountry levels for the feature, i.e. for admin 2, country level, admin1 level and admin2 level names and ids will all be included. The names for these columns are taken from the configured geojson property names
- mean and sd columns for each indicator, except category-type indicators, where only the category names are included as values, so they only have a single column

A toast-like snackbar appears if there is any error with downloading the file. 

This is a pretty simple user interaction - we might let user set the file name or make other choices about the download in future. 

It doesn't show any spinner or progress indicator while the file is being generated - it's generally quick enough not to be necessary I think, but possibly borderline for Brazil?

Key code changes:
- a new composable, `useExcelDownload` which implements composing and writing the file, and sets a `Ref<null | Error>` on itself if any error occurs. Uses the "xlsx" package (as we did in wodin) - the latest versions of this aren't on npm any more, so using their cdn to install. 
- `ExcelDownloadButton` component which shows as a small icon button, and displays the snackbar on error. Maybe not obvious enough in the UI? Let me know if you have any better ideas!


public/dengue/may24/resources/config.json
```diff
@@ -3,8 +3,10 @@
     "countries": ["AGO","ARG","ATG","AUS","BDI","BEN","BFA","BGD","BHS","BLZ","BOL","BRA","BRB","BTN","BWA","CAF","CHN","CIV","CMR","COD","COG","COL","COM","CPV","CRI","CUB","DJI","DOM","ECU","ERI","ETH","FJI","GAB","GHA","GIN","GLP","GMB","GNB","GNQ","GRD","GTM","GUF","GUY","HND","HTI","IDN","IND","JAM","KEN","KHM","LAO","LBR","LCA","LKA","MDG","MEX","MLI","MMR","MOZ","MRT","MTQ","MUS","MWI","MYS","MYT","NAM","NCL","NER","NGA","NIC","NPL","OMN","PAN","PER","PHL","PNG","PRI","PRY","PYF","REU","RWA","SDN","SEN","SGP","SLB","SLE","SLV","SOM","SSD","SUR","SYC","TCD","TGO","THA","TLS","TON","TTO","TWN","TZA","UGA","URY","VCT","VEN","VIR","VNM","VUT","WSM","YEM","ZMB","ZWE"],
     "countriesWithoutAdmin2": ["ATG","BHS","BLZ","BRB","COM","CPV","GRD","JAM","LCA","MUS","MYT","PRI","PYF","SGP","SYC","TTO","VCT"],
     "geoJsonFeatureProperties": {
+        "idAdm0": "GID_0",
         "idAdm1": "GID_1",
         "idAdm2": "GID_2",
+        "nameAdm0": "COUNTRY",
```
EmmaLRussell, line 9: The country (level 0) props defined here aren't actually extracted from the features - it turned out that we don't get the country name in the features (they are removed to save space by Wes's script), and anyway it seemed inefficient to pull out the country name from every feature in a country. 

So instead, we get the country name from the new countryNames.json resource, but we do use these props to name the columns in the excel file.

src/components/ExcelDownloadButton.vue
```diff
@@ -0,0 +1,33 @@
...
+
+watch(downloadError, () => {
+    snackbarOpen.value = !!downloadError.value;
+});
+</script>
+
+<style lang="scss">
+.v-snackbar__wrapper {
+    background-color: rgba(255, 255, 255, 1) !important;
+}
```
EmmaLRussell, line 32: Snackbar showed up as annoyingly transparent without this, maybe there's a more elegant way to address it, but setting the opacity on the component didn't seem to work for me...

src/composables/useExcelDownload.ts
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        XLSX.writeFile(workbook, fileName);
+    };
+
+    const download = () => {
+        downloadError.value = null;
+        debounce(() => {
+            try {
+                downloadFile();
```
EmmaLRussell, line 129: Try throwing an error here to see the snackbar in action..

public/dengue/may24/resources/config.json
M-Kusumgar: ohhh niceee, i actually needed this in my map settings menu PR haha, will probably pull this one in once it is merged to main!

src/components/ExcelDownloadButton.vue
M-Kusumgar: i would have just done it as `variant="elevated" color="#a50026"` (example colour that i tried but any colour really), i believe the opacity didnt work because the actual color for `"error"` is itself faded

M-Kusumgar (approved):
works great, I think this is fairly obvious on the UI especially if we add this into the user instructions!

Couple of things on styling:
* the icons in the top right now look like they are following a different theme (round button instead of rounded squares), i think if possible it would be nice if they were rounded squares and potentially even black background might make them stand out, not sure about that suggestion though!
* i think the tooltips for these icons should be to the left of them, they are correctly at the bottom i think which covers the download button when you hover over help

But also i think these issues can be dealt with in another PR

PR #31, opened 2024-05-14
Mrc 5354 Refactor to change to single update for any user interaction

M-Kusumgar:
Alright so motivation for this was a bug in the admin toggle branch #22. If you zoomed in a little bit on the world view and then clicked on a country, there would be 2 quick updates fired to leaflet (change country, change admin level) and it was showing admin 2 data for country but zoomed out to world view.

The main change in this refactor is that instead of having separate store items like `selectedCountryId` and `selectedAdminLevel` we have `mapSettings` which is an object that mirrors the url (has the same keys).

There were challenges along the way, have put comment for changes that may not make complete sense


src/components/Choropleth.vue
```diff
@@ -34,36 +34,39 @@ import { APP_BASE_ROUTE } from "../router/utils";
...
+const getFeatureProp = (feature: Feature, key: "name" | "id") => {
+    const isAdmin2 = featureInSelectedCountry(feature, mapSettings.value.country) && mapSettings.value.adminLevel === 2;
+    let propertyKey: string;
+    if (key === "id") {
+        propertyKey = isAdmin2 ? featureProperties.idAdm2 : featureProperties.idAdm1;
+    } else {
+        propertyKey = isAdmin2 ? featureProperties.nameAdm2 : featureProperties.nameAdm1;
+    }
+    return feature.properties[propertyKey];
+};
```
M-Kusumgar: inspired by change by @EmmaLRussell  https://github.com/mrc-ide/arbomap/pull/12/files#diff-4c9d4b19a4dd499f75587a26f06db5bf0670247d97ed16a2463b93b5cfb2e790R50-R66, basically we werent accounting for countries with no admin2 data which was leading to issues, the `mapSettings` object now accounts for that and the `adminLevel` inside it is set to 1 for countries without admin 2 data so we can just use that now

src/components/Choropleth.vue
```diff
@@ -86,18 +89,19 @@ const { map, dataSummary, lockBounds, updateLeafletMap, handleMapBoundsUpdated,
     layerOnEvents
 );
 useLoadingSpinner(map, mapLoading);
+const { selectedFeatures } = useSelectedMapInfo();
```
M-Kusumgar: `selectedFeatures` and `selectedIndicators` were getters before so why did I extract them out into a composable?? great question, the watcher was firing as soon as the `mapSetting` changed, **before** the getters for the `selectedFeatures` had fired so the `updateMap` function was getting old `selectedFeatures`, how annoying, but as composables they are fine, getters are probably near the end in the chain of side effects that vue decides to resolve and watchers are probably near the first.

M-Kusumgar:
changes in the leaflet file cost me almost a day to fix, it was nuts, the actual error was about bounds being undefined and nothing about the layers at all, so i tinkered with bounds and then googled a bunch, went down a rabbit hole of module federation because someone had a similar issue with leaflet about this, basically declaring a shared instance of leaflet across different bundles.

eventually all i could do was put console.log everywhere and it all worked, no error was found in the function i changed here, i then commented out the smallest units of leaflet code that i could in the entire file and eventually found commenting out the outline of the country didn't return the error.

i have thought about this for a while and i have no idea, if you get any ideas please do let me know

edit: new info as I was working on admin 1 outlines when admin 2 region selected, it actually isnt anything to do with not adding them instantly, it is actually to do with adding a polyline **before** the geojson, i dont know why! so have created an empty layer and now we can just add layers how we please as long as we have the empty layer in the map

src/composables/useSelectedMapInfo.ts
M-Kusumgar, line 1: should be basically the same as the getters

src/pages/index.vue
```diff
@@ -53,72 +57,67 @@ const props = defineProps({
...
+    if (unknownProps.value.length) return;
+
+    let path = "";
+    path += `/${props.pathogen || PATHOGEN}`;
+    path += `/${props.version || VERSION}`;
+    path += `/${props.indicator || indicatorNames.value[0]}`;
+    const currentPath = router.currentRoute.value.path;
+    if (currentPath.indexOf(path) !== 0) {
+        router.replace(path);
+    }
```
M-Kusumgar: since any number of these props can be empty strings, we construct the url with default values if missing and see if that is a prefix of the current path, if not then we replace the route.

src/pages/index.vue
```diff
@@ -53,72 +57,67 @@ const props = defineProps({
...
-            }
-        } else {
-            // No indicator selected on route - default to first indicator and navigate
-            router.replace(`/${APP_BASE_ROUTE}/${indicatorNames.value[0]}`);
-        }
+    const newMapSettings: MapSettings = {
+        pathogen: checkRouteProp("pathogen", possibleValuesForProps.pathogen),
+        version: checkRouteProp("version", possibleValuesForProps.version),
+        indicator: checkRouteProp("indicator", possibleValuesForProps.indicator),
+        country: checkRouteProp("country", possibleValuesForProps.country),
```
M-Kusumgar: we call `checkRouteProp` again just to get the correct casing

tsconfig.vitest.json
```diff
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "composite": true,
     "tsBuildInfoFile": "./node_modules/.tmp/tsconfig.vitest.tsbuildinfo",
-    "lib": [],
+    "lib": ["DOM"],
```
M-Kusumgar, line 6: HTMLElement was not defined in testing files and was showing a lot of annoying errors

package-lock.json
david-mears-2, line 1: Thanks for your service 🫡

david-mears-2 (commented):
Thank you for your service 🫡

package-lock.json
david-mears-2, line 1: What are the changes to package-lock.json about?

src/composables/useSelectedMapInfo.ts
```diff
@@ -0,0 +1,43 @@
+import { storeToRefs } from "pinia";
+import { useAppStore } from "../stores/appStore";
+
+export const useSelectedMapInfo = () => {
+    const { mapSettings, admin1Indicators, admin2Indicators, admin1Geojson, admin2Geojson } =
+        storeToRefs(useAppStore());
+
+    const selectedIndicators = computed(() => {
+        // get single dictionary of feature id to indicator values
```
david-mears-2, line 10: I really appreciate the explanatory comments

src/components/Choropleth.vue
```diff
@@ -34,43 +34,39 @@ import { APP_BASE_ROUTE } from "../router/utils";
...
 
 const featureInSelectedCountry = (feature: Feature, selectedCountry: string) =>
     feature.properties[featureProperties.country] === selectedCountry;
 
-const featureAdminLevel = (feature: Feature, selectedCountry) => {
-    return featureInSelectedCountry(feature, selectedCountry) &&
-        !appConfig.value.countriesWithoutAdmin2.includes(selectedCountry)
-        ? 2
-        : 1;
+const getFeatureProp = (feature: Feature, key: "name" | "id") => {
```
david-mears-2: Rather than 'getFeatureProp' I prefer 'getFeatureProperty' in line with `featureProperties`, otherwise when I start trying to read the function, my first hypothesis is that its about Vue component props.

src/components/Choropleth.vue
```diff
@@ -34,43 +34,39 @@ import { APP_BASE_ROUTE } from "../router/utils";
...
+    const isAdmin2 = featureInSelectedCountry(feature, mapSettings.value.country) && mapSettings.value.adminLevel === 2;
```
david-mears-2: Is it correct to check whether the feature is in the selected country as part of determining whether it's an admin2 feature? I wouldn't expect that behavior from a function whose job is to look up a property of a feature.

src/utils.ts
```diff
@@ -9,3 +11,15 @@ export function debounce(fn, wait = 0) {
...
+    if (mapSettings1 === null && !!mapSettings2) return false;
+    if (mapSettings2 === null && !!mapSettings1) return false;
+    return (
+        mapSettings1.pathogen === mapSettings2.pathogen &&
+        mapSettings1.version === mapSettings2.version &&
+        mapSettings1.indicator === mapSettings2.indicator &&
+        mapSettings1.country === mapSettings2.country &&
+        mapSettings1.adminLevel === mapSettings2.adminLevel
+    );
+};
```
david-mears-2: I wonder how future coders will remember to add extra checks to this function when they add properties to mapSettings. Can we programmatically infer what properties we need to check? E.g.

```ts
    const keys = Object.keys(mapSettings1) as (keyof MapSettings)[];
    return keys.every(key => mapSettings1[key] === mapSettings2[key]);
```

david-mears-2: That might not be ideal, it's just an example

src/components/Choropleth.vue
EmmaLRussell: I think the downside of this might be that you end up recalculating the `selected..` computeds for each instance of using the composable, whereas the getter would only be calculated once, in the store. Or is Vue smart enough to only calculate once when the parameters are the same? Either way, it solved the problem, and as long as it isn't impacting performance too much...

src/components/Choropleth.vue
```diff
@@ -5,15 +5,15 @@
...
+            @ready="updateMap"
             :max-bounds-viscosity="1"
         >
             <LTileLayer v-once data-testid="tile-layer" v-bind="backgroundLayer"></LTileLayer>
             <LControl position="bottomright">
                 <Legend :numberOfSteps="6" />
             </LControl>
             <LControl position="topleft">
-                <ResetMapButton :selected-indicator="selectedIndicator" @reset-view="updateRegionBounds" />
+                <ResetMapButton :selected-indicator="mapSettings.indicator" @reset-view="updateRegionBounds" />
```
EmmaLRussell, line 16: OK, so this sort of prop works because whenever anything in `mapSettings` updates, the whole reactive object gets replaced...?

src/stores/appStore.ts
```diff
@@ -9,101 +8,63 @@ import {
...
             }
 
+            let level = adminLevel;
             // Some countries do not have admin2 regions or data - if one of these is selected, we load
             // a more detailed geojson, and re-use its admin1 indicators as "admin2"
-            const level = this.appConfig.countriesWithoutAdmin2.includes(countryId) ? 1 : 2;
+            const admin2DataMissing = state.appConfig.countriesWithoutAdmin2.includes(country);
+            if (level === 2 && admin2DataMissing) {
+                level = 1;
+                state.admin2Indicators[country] = state.admin1Indicators[country];
```
david-mears-2, line 53: I see that you didn't introduce this in this PR, but it makes me uneasy to store admin1 indicators under the name 'admin2', for reasons of making how things work more confusing. What is the behavior of the app when someone selects 'admin 2' on a country missing that level - are they shown admin1, and does the toggle stay on admin 1? Maybe I'd prefer that we don't offer the 'admin 2' option to users if they have selected a country without admin 2 regions. Then, we could here update the selected admin level from 2 to 1. If people agree with that, I'll make a ticket for it.

src/components/Choropleth.vue
```diff
@@ -34,43 +34,39 @@ import { APP_BASE_ROUTE } from "../router/utils";
...
-        : 1;
+const getFeatureProp = (feature: Feature, key: "name" | "id") => {
+    const isAdmin2 = featureInSelectedCountry(feature, mapSettings.value.country) && mapSettings.value.adminLevel === 2;
+    let propertyKey: string;
+    if (key === "id") {
+        propertyKey = isAdmin2 ? featureProperties.idAdm2 : featureProperties.idAdm1;
+    } else {
+        propertyKey = isAdmin2 ? featureProperties.nameAdm2 : featureProperties.nameAdm1;
+    }
+    return feature.properties[propertyKey];
```
EmmaLRussell: OK, you're saying that we don't need to check `countriesWithoutAdmin2` here any more, because the mapSetting.adminLevel won't be set to 2 for countries without admin2, hence the refactor, removing `featureAdminLevel`. To be honest, I don't reallly see the benefit of combining id and name getters into one method - it's not really much terser and it feels less readable to me. Subjective perhaps though!

src/composables/useSelectedMapInfo.ts
M-Kusumgar: these were already there by you guys actually just kept them :laughing:

package-lock.json
M-Kusumgar: i should revert this actually, there was a rabbit hole about using vite plugins so i installed and uninstalled a couple of things, will revert!

src/stores/appStore.ts
```diff
@@ -9,101 +8,63 @@ import {
...
-                }
+            if (level === 2 && !admin2DataMissing && !(country in state.admin2Indicators)) {
+                state.admin2Indicators[country] = await getIndicators(country, level);
             }
-            if (!(countryId in this.admin2Geojson)) {
-                this.admin2Geojson[countryId] = await getGeojsonFeatures(countryId, level);
+
+            if (!(country in state.admin2Geojson)) {
+                state.admin2Geojson[country] = await getGeojsonFeatures(country, level);
             }
```
david-mears-2:
```suggestion
state.admin2Geojson[country] = state.admin2Geojson[country] || await getGeojsonFeatures(country, level);
```

This achieves the same thing and is terser, but maybe it's less readable? (If people like the terse option, it can apply a few lines back, too.)

src/components/Choropleth.vue

src/composables/useColourScale.ts
```diff
@@ -17,7 +18,7 @@ type FillAndOutlineColour = {
     outlineColor: string;
 };
 
-export const useColourScale = (selectedIndicators: Ref<Dict<FeatureIndicatorValues>>) => {
+export const useColourScale = () => {
```
EmmaLRussell: There was a bit of a discussion in a previous branch about whether in general we should be avoiding composables from using eachother and getting the client component to coordinate them, and pass in params to the composables as required. Feels like it may make them more easily testable, and cleaner separation - but I don't feel too strongly about it..

src/stores/appStore.ts
```diff
@@ -9,101 +8,63 @@ import {
...
 
-            this.admin0GeojsonFeature = (await getGeojsonFeatures(countryId, 0))[0];
-            this.selectedCountryId = countryId;
+            state.admin0GeojsonFeature = (await getGeojsonFeatures(country, 0))[0];
+            // need to account for the fact that some countries do not have admin level 2 data
```
david-mears-2: Is this a TODO comment, or what does it mean?

src/components/Choropleth.vue
M-Kusumgar: i think this is in the correct place, only because out of visible features, lets say an admin 2 feature will have props `ID_1`, `ID_2` which are admin level 1 and 2 ids respectively, since the feature itself is an admin level 2 feature, even though it store admin level 1 id, we actually want the admin level 2 id, that is its own id

src/composables/useLeaflet.ts
```diff
@@ -134,8 +135,14 @@ export const useLeaflet = (
 
         resetMaxBoundsAndZoom();
 
+        // please keep addition of empty layer first
+        // more info: https://github.com/mrc-ide/arbomap/pull/31#issuecomment-2110480157
```
EmmaLRussell: oh wow 😱

src/composables/useSelectedMapInfo.ts
```diff
@@ -0,0 +1,43 @@
...
+
+        // Return admin1 indicators for countries other than the selected country, and
+        // admin2 indicators for the selected country
```
EmmaLRussell, line 13: Might be worth modifying this comment a little to cover the case where a country may be selected, but mapSettings.adminLevel is 1 because the country does not have admin2 - in which case it will just default to returning all admin1 (if I understand correctly).

src/utils.ts
M-Kusumgar: yhhh i like this! there is something to be said about using one of them as the source for the keys but assuming they are both typed correctly as `MapSettings` this is nice

src/composables/useLeaflet.ts
david-mears-2: This links to a comment that I don't actually understand. Can you summarise why we need an empty layer?

src/pages/index.vue
```diff
@@ -42,72 +46,67 @@ const props = defineProps({
...
-);
-
-watch([countryToSelect, admin1Indicators], async () => {
-    // wait until admin1 loaded before selecting country
-    if (countryToSelect.value !== null && Object.keys(admin1Indicators.value).length) {
-        await selectCountry(countryToSelect.value);
-        countryToSelect.value = null;
-    }
-});
+watch([() => props, initialisationComplete], selectDataForRoute, { deep: true });
```
EmmaLRussell, line 107: You need to deep watch for the props?

src/components/Choropleth.vue
david-mears-2: But why does that mean we need to check the feature is part of the selected country?

src/utils.ts
david-mears-2: Maybe it's possible to iterate over the keys of MapSettings rather than over the keys of one of the instances.

src/composables/useLeaflet.ts
david-mears-2: (I know that it's fundamentally mysterious, but can you summarise the part that isn't mysterious?)

src/stores/appStore.ts
M-Kusumgar: The selected admin level is `1` if a country has missing admin level 2 data, that is what the `finalMapSettings` line does in this function so i think in the future when the admin level toggle is added it will show admin level 1 features only, but i agree that there could be a potentially better method to doing this than putting the admin 1 indicators into admin 2 indicators

src/stores/appStore.ts
M-Kusumgar: im happy to do that, i think it is readable but lets confirm with @EmmaLRussell

src/stores/appStore.ts
M-Kusumgar: that comment was supposed to explain the next line, i change the admin level in the `finalMapSettings` because i am accounting for the countries that do not have admin level 2 data, happy to change the comment so it makes more sense, any suggestions?

src/pages/index.vue
```diff
@@ -1,9 +1,11 @@
-<template>
+<template v-if="initialisationComplete">
```
david-mears-2, line 1: Just checking, would this prevent the spinner from appearing while things are being initialised?

src/pages/index.vue
EmmaLRussell: :(

src/stores/appStore.ts
david-mears-2: Actually, terser still is using OR assignment:

```ts
state.admin2Geojson[country] ||= await getGeojsonFeatures(country, level);
```

src/stores/appStore.ts
david-mears-2: OK. If you just drop `need to` then I understand it as an explanation.

src/pages/index.vue
EmmaLRussell: I find this refactor quite confusing. I don't really get the benefit of building the "base path OR requested path" WITH the assumption that either is fine because you've already checked for unknowns. I get that the nesting in the previous implementation was  gross, but at least it was reasonably clear what case you were dealing with in each arm - this just seems to attenuate the confusion.. And then having to check values a second time to sort out the casing doesn't seem optimal either.

I don't have any brilliant suggestions for what to replace it with. Maybe just some more comments would help. But it would definitely be good to avoid the double check if possible.

src/pages/index.vue
```diff
@@ -42,72 +46,67 @@ const props = defineProps({
...
-    checkRouteProp("version", [VERSION]);
-    const indicator = checkRouteProp("indicator", Object.keys(appConfig.value.indicators));
-    const country = checkRouteProp("country", appConfig.value.countries);
+    Object.keys(possibleValuesForProps).forEach((prop: PropName) => {
+        if (checkRouteProp(prop, possibleValuesForProps[prop]) === undefined) {
+            unknown.push(prop);
+        }
+    });
 
     unknownProps.value = unknown;
```
david-mears-2, line 86: Not really your job in this PR, but maybe we can fold in a renaming of 'unknown' since my brain was trying to read this as the reserved word referring to a type, whereas `unknown` is an array.

src/types/storeTypes.ts
```diff
@@ -1,21 +1,21 @@
...
+    // these are initialised as empty objects and extended every time the user selects a
+    // country
     admin2Indicators: Dict<FeatureIndicators>;
-    admin2Geojson: Dict<Geojson>;
-    admin0GeojsonFeature: Feature | null;
+    admin2Geojson: Dict<MapFeature[]>;
+
+    // these variables change every time there is an update to the map
+    mapSettings: MapSettings | null;
+    admin0GeojsonFeature: MapFeature | null;
```
EmmaLRussell, line 20: Have the feature dictionaries actually changed, or were these just wrongly defined before?

src/stores/appStore.ts
```diff
@@ -9,101 +8,63 @@ import {
...
-                    this.admin2Indicators[countryId] = await getIndicators(countryId, level);
-                }
+            if (level === 2 && !admin2DataMissing && !(country in state.admin2Indicators)) {
+                state.admin2Indicators[country] = await getIndicators(country, level);
             }
-            if (!(countryId in this.admin2Geojson)) {
-                this.admin2Geojson[countryId] = await getGeojsonFeatures(countryId, level);
+
+            if (!(country in state.admin2Geojson)) {
+                state.admin2Geojson[country] = await getGeojsonFeatures(country, level);
```
EmmaLRussell: So this is *actually* going to load the more detailed admin1 geojson for countries without admin 2 , and will store it in `admin2Geojson`? And something similar is happening for the indicators - it copies the admin1 indicators into `admin2Indicators` for that country. Bit confusing. I wonder if, in another branch, we should rename `admin1` and `admin2` indicators and geojson to something like `selectedCountryIndicators` etc and ... uh.. `unSelectedCountries...`.. 😆 (but not that). "detailed" and "simple"..?

EmmaLRussell (commented):
This is actually doesn't *too* disruptive to me - most of the changes for most users of the store are pretty simple. 👍 

Mostly questions, but generally LGTM! I wasn't able to break it!

src/components/Choropleth.vue
david-mears-2: > I don't reallly see the benefit of combining id and name getters into one method

:heavy_plus_sign:

src/composables/useColourScale.ts
david-mears-2: I'm easy / not experienced enough to have a view on this

src/stores/appStore.ts
david-mears-2: Agree that it's perverse to store admin2 info under the name admin1. I don't know why we do that.

src/pages/index.vue
```diff
@@ -42,72 +46,67 @@ const props = defineProps({
...
-const notFoundMsg = (valueType, value) => `Unknown ${valueType}: ${value}.`;
+const notFoundMsg = (valueType: string, value: string) => `Unknown ${valueType}: ${value}.`;
 
 const notFoundDetail = computed(() => {
     return unknownProps.value.map((propName) => notFoundMsg(propName, props[propName])).join(" ");
 });
 
-const countryToSelect: Ref<null | string> = ref(null);
+const checkRouteProp = (propName: PropName, candidates: string[]) => {
+    // Do case-insensitive check against route prop - add to unknown props if not found
```
david-mears-2: We no longer seem to add to unknown so comment is outdated

src/pages/index.vue
```diff
@@ -42,72 +46,67 @@ const props = defineProps({
...
-        const result = candidates.find((i) => pattern.test(i));
-        if (!result) {
-            unknown.push(propName);
-        }
-        return result;
+    const possibleValuesForProps: Record<PropName, string[]> = {
+        pathogen: [PATHOGEN],
+        version: [VERSION],
+        indicator: Object.keys(appConfig.value.indicators),
+        country: appConfig.value.countries
```
david-mears-2, line 74: I feel like this kind of refactor is unrelated to the main thing this PR is about, is that right? It would make diffs simpler to read if we used separate PRs for separate refactors. Also, if there's pushback on the unrelated refactor it might delay the merging of the main thing the PR is about.

david-mears-2: But I recognise that sometimes it's hard to separate out "two" refactors if you've worked on them concurrently on the same branch.

tests/unit/components/legend.spec.ts
```diff
@@ -19,7 +19,7 @@ const appConfig = {
             unit: "%"
         }
     }
-};
+} as any;
```
david-mears-2, line 22: Hm, should we enforce an AppState type here?

src/components/Choropleth.vue
M-Kusumgar: i think you're right about the computed for each instance of composable use, i couldnt find anything online about vue optimisation, i think the performance is alright but we should probably address this so it doesn't scale, i think there is a very easy way to address this related to your comment about composables not using other composables, everything that uses these computed is in the `Choropleth` so ill just pass these refs into these other composables! should fix performance concerns

src/components/Choropleth.vue
M-Kusumgar: yep exactly

tests/unit/composables/useSelectedMapInfo.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+                FOI: { mean: 0.3, sd: 0.03 },
+                serop9: { mean: 0.4, sd: 0.04 },
+                hosp_total: { mean: 0.5, sd: 0.05 },
+                hosp_0_4: { mean: 0.6, sd: 0.06 },
+                hosp_5_9: { mean: 0.7, sd: 0.07 }
+            }
+        });
+    });
+
+    test("selectedFeatures includes selected country's admin2 features, and admin1 for all others", async () => {
```
david-mears-2:
```suggestion
    test("selectedIndicators includes selected country's xyz, and xyz", async () => {
```

tests/unit/composables/useSelectedMapInfo.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+        const store = useAppStore();
+        await store.initialiseData();
+        await store.updateMapSettings(mockMapSettings({ country: "TZA", adminLevel: 2 }));
+        const { selectedFeatures } = useSelectedMapInfo();
+        expect(selectedFeatures.value).toStrictEqual([
+            { properties: { shapeID: "789-a", shapeName: "Test789-a", shapeGroup: "TZA" } },
+            { properties: { shapeID: "789-b", shapeName: "Test789-b", shapeGroup: "TZA" } },
+            { properties: { shapeID: "123", shapeName: "Test123", shapeGroup: "MWI" } }
+        ]);
+    });
```
david-mears-2, line 32: We should probably test the case for selectedFeatures where a country is selected and the admin level selected is 1.

tests/unit/composables/useSelectedMapInfo.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+
+describe("useSelectedMapInfo", () => {
+    beforeEach(() => {
+        setActivePinia(createPinia());
+    });
+
+    test("selectedFeatures includes all admin 1 features when there is no selected country", async () => {
+        const store = useAppStore();
+        await store.initialiseData();
+        await store.updateMapSettings(mockMapSettings());
```
david-mears-2: Can you set adminlevel to something other than 1, to make this test slightly more meaningful? (And to help it withstand changes in mockPinia.ts)

tests/unit/composables/useSelectedMapInfo.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+            { properties: { shapeID: "789-a", shapeName: "Test789-a", shapeGroup: "TZA" } },
+            { properties: { shapeID: "789-b", shapeName: "Test789-b", shapeGroup: "TZA" } },
+            { properties: { shapeID: "123", shapeName: "Test123", shapeGroup: "MWI" } }
+        ]);
+    });
+
+    test("selectedIndicators includes all admin 1 features when there is no selected country", async () => {
+        const store = useAppStore();
+        await store.initialiseData();
+        await store.updateMapSettings(mockMapSettings());
```
david-mears-2: Can you set adminlevel to something other than 1, to make this test slightly more meaningful? (And to help it withstand changes in mockPinia.ts)

tests/unit/composables/useSelectedMapInfo.spec.ts
david-mears-2: Same goes for selectedIndicators.

src/components/Choropleth.vue
M-Kusumgar: update on this, have only kept the composable in choropleth, legend vue files and useDataSummary composable, passed it in everywhere else, should scale a bit better

tests/unit/pages/index.spec.ts
```diff
@@ -1,14 +1,14 @@
 import { render, screen } from "@testing-library/vue";
 import { describe, expect, test, vi, beforeEach, afterAll } from "vitest";
 import Index from "@/pages/index.vue";
-import router from "@/router";
 import { flushPromises } from "@vue/test-utils";
+import router from "../../../src/router";
```
david-mears-2: Why do you prefer the relative path over the absolute one with @?

src/components/Choropleth.vue
M-Kusumgar: yh i kinda agree, it just felt like the two functions were doing the same thing, i have a better idea for how to make this function cos i realised that we control feature properties, have had another go at the function, let me know what you think, if its still less readable, happy to revert back to the old functions

M-Kusumgar: also yes we should not have to worry about `countriesWithoutAdmin2` that kind of consideration should be made in only one place i think, otherwise it is easy to forget, and we already do it in the store

src/composables/useColourScale.ts
M-Kusumgar: yep have reverted this, because of this concern and performance concerns that come with sticking the composable everywhere

src/composables/useLeaflet.ts
M-Kusumgar: yh... what a journey

src/composables/useSelectedMapInfo.ts
M-Kusumgar: yep exactly, have added comment

src/pages/index.vue
M-Kusumgar: yep shouldnt have to change this line if we add more props (e.g. `adminLevel` soon) or ever again XD

tests/unit/stores/appStore.spec.ts
```diff
@@ -24,59 +25,18 @@ describe("appStore", () => {
             expect(store.countryBoundingBoxes).toStrictEqual(MOCK_BOUNDS);
             expect(store.admin2Indicators).toStrictEqual({});
             expect(store.admin2Geojson).toStrictEqual({});
-            expect(store.selectedIndicator).toBe("");
-            expect(store.selectedCountryId).toBe("");
+            expect(store.initialisationComplete).toStrictEqual(true);
         });
 
         test("selectCountry selects country and loads data", async () => {
```
david-mears-2: This needs a new description

david-mears-2 (changes requested):

src/pages/index.vue
M-Kusumgar: i have found a way to avoid the double check by just storing the checked versions in an object, will have a think about this path stuff, the thing is i found the previous implementation much more confusing because it just happened to work, what i mean by that is it was hinging on whether or not the indicator was empty, but it would not work if we had another disease or version of data, this type of logic will work for different diseases and versions

src/types/storeTypes.ts
M-Kusumgar: i havent changed any structure of these, they were just wrongly defined i think, my IDE makes the file title red and red squiggles everywhere if it is the wrong type, does your IDE not do that?

src/stores/appStore.ts
M-Kusumgar: haha maybe yh, would be nice to have these names make more sense!

src/composables/useLeaflet.ts
M-Kusumgar: in summary:
* if we try to add a polyline after a geojson to a map with no layers it errors
* if we add a geojson after a polyline to a map it is fine but that means the country outline would be underneath the geojson and it looks ugly
* if we add a polyline to the map first and then add any layer afterwards it is also fine

I think it must be some sort of weird bug with the implementation of polyline so i have chosen a polyline for the empty layer and this layer will never be removed from the map and will only be added once

src/components/Choropleth.vue
M-Kusumgar: because we have no way of knowing if the feature is an admin 1 or admin 2 feature, however we do have the knowledge that if a country is **not** selected then the feature must be admin level 1 feature, so if it is selected and the state says admin level 2 then we know that this is an admin level 2 feature

src/utils.ts
M-Kusumgar: pretty sure that is not possible, since types are stripped away when typescript is transpiled into base javascript, we cant have any javascript implementation dependent on a type

src/pages/index.vue
M-Kusumgar: yes i believe it does prevent the initial spinner, can make a ticket if that is a problem, seemed pretty fast to me though

src/pages/index.vue
M-Kusumgar: yes would definitely second this

src/pages/index.vue
M-Kusumgar: removed that bit

src/pages/index.vue
M-Kusumgar: i had to change a bit of this function for the refactor but i found the implementation confusing as i was changing the relevant bits of the refactor so thought i should just change the whole function because i actually think the function was doing something incorrectly anyway, i know we dont have different diseases/versions yet but this function would not have worked for those cases

but yh shouldve probably left it to another smaller PR, will see, if there is much pushback to it, ill revert it to older function

tests/unit/components/legend.spec.ts
M-Kusumgar: i dont personally think it is necesasary considering this is just a test, its actually a typescript problem in my opinion, it is complaining that we do not have `description`s in the indicators object but it would be nice if there was a way for  `Partial<>` type to be recursively applied since we dont need `description`s for the purposes of the test

tests/unit/composables/useSelectedMapInfo.spec.ts
M-Kusumgar: nice spot

tests/unit/composables/useSelectedMapInfo.spec.ts
M-Kusumgar: the admin level should never be anything other than 1 if there is no selected country, same for indicator comment below

tests/unit/pages/index.spec.ts
M-Kusumgar: my IDE was complaining about this, i dont know why but just changed it to relative and it fixed it

src/components/Choropleth.vue
david-mears-2: Nice one :+1:

tests/unit/stores/appStore.spec.ts
M-Kusumgar: done

tests/unit/composables/useSelectedMapInfo.spec.ts
M-Kusumgar: done

src/stores/appStore.ts
EmmaLRussell: I'd vote for the original suggestion personally, not a huge fan of OR assignment, though it's really just aesthetic preference at this point.

src/components/Choropleth.vue
david-mears-2: [I've just clocked it wasn't you who introduced this logic in the first place so sorry for jumping on it!]

I _think_ I follow... though you've just described the proxy we're using as checking "if a country is **not** selected" rather than checking if the feature is "in" the selected country.

So it seems we have two disadvantages here to my mind, one is that the way this logic works is non-obvious, and the other is that we are relying on the 'knowledge' that "if a country is **not** selected then the feature must be admin level 1 feature". Since that isn't a law of physics it's possible that it won't always be true. For example, maybe at some point we introduce an admin0-level view; or something changes that means that the `featureInSelectedCountry` check is no longer a totally accurate proxy for 'a country is selected' (could it happen around disputed regions? I don't know, but I'd rather not be forced to reason about it.)

An attempt to short-cut the two disadvantages: check whether the feature is admin2 by seeing whether it has the `featureProperties.idAdm2` key. (This scales nicely if we introduce admin level 3!)

src/utils.ts
david-mears-2: Good to know!

src/composables/useLeaflet.ts
david-mears-2: Thanks! Can you link to here instead of to the longer comment? Which, for posterity, was https://github.com/mrc-ide/arbomap/pull/31#issuecomment-2110480157

src/pages/index.vue
david-mears-2: Ah ok, so it's only waiting for appstore to initialise, not for choropleth and leaflet etc?

src/components/Choropleth.vue
M-Kusumgar: i can potentially see this working, although havent thought about all the implications of this, happy to create a ticket about this because i think your idea will work!

tests/unit/composables/useSelectedMapInfo.spec.ts
david-mears-2: What if we one day introduce an admin-level-0 view?

src/stores/appStore.ts
M-Kusumgar: can do the original suggestion then! done

src/composables/useLeaflet.ts
M-Kusumgar: done

src/components/Choropleth.vue
david-mears-2: If _that_ doesn't work I would advocate for a comment on how checking presence of country is required to diagnose the admin level, eg:
```
// There is no way to tell if a given feature's level is admin1 or admin2. But, if no country is selected, we can infer that the level is admin1.
```

tests/unit/composables/useSelectedMapInfo.spec.ts
M-Kusumgar: im happy to add `adminLevel: 0` or something like that but feels like it isnt closely mocking what actually happens in the execution of the app right now, like feels weird to test a case which isnt currently even possible

src/components/Choropleth.vue
M-Kusumgar: yep made a ticket that links to this discussion! https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-5376/Improve-way-we-get-feature-properties

tests/unit/composables/useSelectedMapInfo.spec.ts
david-mears-2: thanks!

tests/unit/composables/useSelectedMapInfo.spec.ts
david-mears-2: That makes sense. I guess I felt like I wanted the test to prove to me that selectedFeatures wasn't just using the selected admin level to determine which level features it includes.

tests/unit/pages/index.spec.ts
david-mears-2: Can you live with a red underline on this? I prefer the @ one because it won't have to change as much.

tests/unit/composables/useSelectedMapInfo.spec.ts
M-Kusumgar: have added it in :laughing: lets just leave it in for now

tests/unit/pages/index.spec.ts
M-Kusumgar: sure have reverted it

src/components/Choropleth.vue
david-mears-2: I respect the aversion to duplication, but also a little is fine now and then. https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

david-mears-2 (approved):
Thanks for addressing every little thing!!!

src/components/Choropleth.vue
M-Kusumgar: fair enough! have reverted it back (with a tiny change)

src/pages/index.vue
M-Kusumgar: i have thought about this and its a tricky problem, for now i have reverted it to similar path logic to old route checking but have created a [ticket](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-5379/Improve-route-validation) for it since this is a problem we havent solved, would be nice to do it properly at some point in the future

EmmaLRussell: Fair enough - yeah, feels like it should be easy enough to find a nice of expressing this bit of pure logic, but not so!

src/types/storeTypes.ts
EmmaLRussell: It does but it sometimes gets it wrong, so I don't always pay much attention to it! 😆

src/stores/appStore.ts
EmmaLRussell: I think this might have been my fault. Mea culpa. Let me know if you think of good replacement names and we can do that in another ticket. It is basically just "features to show for the selected country" and "features to show for countries which aren't selected" (and indicators).

EmmaLRussell (approved):
Great stuff! :)

PR #32, opened 2024-05-14
mrc-5350 Show app version

EmmaLRussell:
This branch adds some version visibility into the app. 

At the bottom of the text of the About page (but above the footnote), it shows the version from package.json, with a text button to click to show/hide details - these details are the Git branch and short commit, so of no interest to most users. 

There is a new "prebuild" npm step which gets the git details and package version, and writes these out to `src/version.ts` (which is gitignored), where they are available to the page. This step is run as the first part of `build` and `dev`.

This approach will rely on us remembering to update the package version regularly (minor version for new feature, patch for bug fix) - I'd prefer to avoid the annoyance of the NEWS descriptions that we had in HINT - but with the fallback of being able to check git details if need be.

### DIFF
--- a/.github/workflows/deploy-to-ghp.yml
+++ b/.github/workflows/deploy-to-ghp.yml
@@ -4,7 +4,7 @@ name: Deploy static content to Pages
 on:
   # Runs on pushes targeting the default branch
   push:
-    branches: ['main', 'mrc-5207']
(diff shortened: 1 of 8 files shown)


M-Kusumgar (approved):
Niceeee

david-mears-2 (approved):
Thanks for including the branch!

PR #33, opened 2024-05-17
mrc-5199 add no data tooltip

M-Kusumgar:
tooltips in areas without data such as the empty regions in australia were just empty before, now they will still display the name of the region and have no data text in grey


david-mears-2 (approved):

EmmaLRussell (approved):
Great, thanks! :+1:

PR #34, opened 2024-05-19
mrc-5386 Change occurrences of "colour" to "color"

EmmaLRussell:
To be consistent with ourselves, and with css, rename all config and code object with "colour" in their names to "color" instead. 

I've generally left free text - comments and test names - with "colour".


david-mears-2 (approved):
Colour me satisfied!

PR #35, opened 2024-05-20
mrc-5385 Support categorical indicators for serop9 classification

EmmaLRussell:
Implements a new indicator requested by WHO, which categorises the serop9 values into high/medium/low and shows these categories as three colours only on the map view.

The levels are Under 40%, 40%-60% and Above 60%

Clare suggested red/orange/yellow for these categories (red for low = inappropriate for vaccine intervention). 

image

The easiest way to do this seemed to be to copy the serop9 values into a new indicator field, `serop9_class` and make a new indicator config for it. Hence the updates to the R script and indicator files, 

The code changes support a new type of colour view, by category rather than scale. Rather than checking where in a normalised scale an indicator value falls, and calling a colour method to find the correct colour for the value, we need to check which of the configured categories it falls into. 

The indicator config changes are:
- replace `colorScale` with `colors` which has `type`: `scale` or `category`. 
- if type is `scale`, there will be a `colorScale` property
- if type is `category`, there will be an array of `categories`, each of which defines a `name`, `color` and `upperLimit`. The code assumes that these categories are in the correct order, low to high, and the highest category can have `null` as its upper limit. 

Key code changes:
- refactored the `useColorScale` to `useIndicatorColors` and replaced the computed `colorScales` property with helper methods to support the needs of the Legend and ColorScaleIcon as well as the Choropleth and useTooltips composable. The idea is that clients shouldn't need to call the colourScale method itself when getting the color for a value, but can let the composable deal with what type of colours the indicator uses. However, it's helpful for the Legend and ColorScale icon to be able to check color type and get the names and colors for the categories, hence the other helper methods.
- Updates to ColorScaleIcon and Legend. Fiddled with the Legend css as category names are a bit longer than the numeric values we've been showing so we don't want that to be fixed width really. I also opportunistically put a background on the Legend since I was changing that css anyway. 
- Updates to useTooltips - we show the category for the new indicator rather than the numeric value.


src/components/indicatorMenu/ColorScaleIcon.vue
```diff
@@ -23,8 +24,18 @@ const props = defineProps({
...
+const iconColors = computed(() => {
+    if (getIndicatorColorType(props.indicator) === ColorType.Scale) {
+        const colorScaleSampleValues = [0.7, 0.9, 0.4, 0.1];
+        return colorScaleSampleValues.map((value) => getIndicatorValueColor(props.indicator, value, false));
+    }
+    const result = [];
+    const categoryColors = getIndicatorColorCategories(props.indicator).map((cat) => cat.color);
+    while (result.length < 4) {
+        result.push(...categoryColors);
+    }
```
M-Kusumgar, line 41: so this is required as there are 3 categories and 4 icon sections? if something happens in the future and `categoryColors` happens to be an empty array this will be impossible to debug, im not too fussed about this because that case is unlikely but there are some easy alternatives, like just repeating `categoryColors` 4 times or something like that, optional suggestion though

M-Kusumgar (approved):
Looks good to me in terms of code, i do slightly dislike how this makes the map look but thats personal preference i think. since it is large blocks of the same color it now does actually look like a jigsaw puzzle with the boundaries but this is what they want i guess!

optional suggestion in a comment but not too fussed about it

src/components/indicatorMenu/ColorScaleIcon.vue
EmmaLRussell: I'll put a check in and throw an error if length is 0 - we're already making some assumptions that the colour categories are configured sensibly, but you're right, this would be a particularly bad one to hit!

PR #36, opened 2024-05-22
mrc-5398 Bug fix for direct browse to admin1

EmmaLRussell:
Direct browsing to urls like `/dengue/may24/FOI/VEN/admin1` were failing because the store was wrongly attempting to load detailed (aka "admin2", misleadingly) geojson for the selected country even if the non-detailed admin1 level was selected. However it was also choosing the filename to attempt to load based on the selected level, admin1. Hence attempting to load detailed admin1 files for countries which did not have them i.e. countries which do have admin2 data (most of them). 

Moving the detailed geojson load inside the check for initial requested admin level == 2 fixes this problem. 

This code is pretty confusing and sometimes misleading - we're going to refactor it in a future ticket when we revamp the geojson loading. 

I *was* able to reproduce the bug for Brazil after I did a browser restart so maybe it wasn't showing before for us because of store being retained, but it all seems a bit of a coincidence... 🤷

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "dengue-map",
-  "version": "0.2.1",
+  "version": "0.3.1",
   "type": "module",
   "engines": {
     "node": "^20.0.0"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "dengue-map",
-  "version": "0.2.0",
+  "version": "0.2.1",
   "type": "module",
   "engines": {
     "node": "^20.0.0"
--- a/tests/e2e/routing.spec.ts
+++ b/tests/e2e/routing.spec.ts
@@ -63,6 +63,16 @@ test.describe("Router", () => {
         await expectIndexPage(page, "/FOI/TZA", "FOI", "Force of infection", "TZA", "interpolateRdYlBu", 2070, 186);
     });
 
+    test("browse to indicator and country at admin1 loads expected data", async ({ page }) => {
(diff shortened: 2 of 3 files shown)


M-Kusumgar (approved):
couldnt get the same bug in this branch so works for me!

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "dengue-map",
-  "version": "0.2.0",
+  "version": "0.2.1",
```
david-mears-2:
```suggestion
  "version": "0.3.1",
```

we are on 0.3.0 now

tests/e2e/routing.spec.ts
```diff
@@ -63,6 +63,16 @@ test.describe("Router", () => {
...
 
+    test("browse to indicator and country at admin1 loads expected data", async ({ page }) => {
+        await page.goto(`${BASE_URL}/FOI/VEN/admin1`);
+        await expect(await page.textContent(".indicator-menu-activator")).toBe("Force of infection");
+        const summary = await page.locator(".choropleth-data-summary");
+        await expect(await summary).toHaveAttribute("color-scale", "interpolateRdYlBu");
+        await expect(await summary).toHaveAttribute("selected-country-id", "VEN");
+        await expect(await summary).toHaveAttribute("feature-count", "1915");
+        await expect(await summary).toHaveAttribute("selected-country-feature-count", "25");
+    });
```
david-mears-2, line 74: This test successfully fails on main :+1:

david-mears-2 (approved):
Sorry, forgot to batch my comments

PR #37, opened 2024-05-29
mrc-5410 Update FOI colour scale

EmmaLRussell:
As requested by science team, updated the FOI colour scale to a single-hue scale.


david-mears-2 (approved):
Looks good!

PR #38, opened 2024-05-30
mrc-5213 Analytics

EmmaLRussell:
This branch adds analytics recording with [umami](https://umami.is/).
The tag to load the analytics script is only included in `index.html` if the current host is production, so no events should be recorded on localhost or the staging github pages site.

This branch is currently deployed to production for testing. 

I've added the email address and password for logging into our account on https://umami.is/ to the vault (secret name "umami") - so you should be able to log in there, then visit pages on https://arbomap.org and see the events appear on the umami dashboard (the Realtime view is easiest for that).

### DIFF
--- a/index.html
+++ b/index.html
@@ -6,8 +6,15 @@
     <link rel="icon" href="/favicon.ico" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title></title>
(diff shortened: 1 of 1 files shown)


M-Kusumgar (approved):
LGTM, pretty fun playing around with it actually, realtime view seemed to work for me!

PR #40, opened 2024-05-30
Mrc 5388 map settings dropdown with country, indicator and admin toggle

M-Kusumgar:
Should be ready to look at! Main differences include:
* the menu is now inside the choropleth (i think this makes more sense anyway, it used to be in the index page)
* on smaller screens it is outside the map, bigger screens it is inside as an `LControl`
* better types
* new `MapSettingsMenu`,  `MenuActivator`, `MenuLayout` components
* also applied to unselected opacity trick to the slide items in the indicator to make the ones that are not selected slightly faded

Do try and test this out with long country names, small mobile screens, the admin toggle should also wrap under the country name when the country name is long!


playwright.config.ts
```diff
@@ -4,7 +4,7 @@ import { defineConfig, devices } from "@playwright/test";
 export default defineConfig({
     testDir: "./tests/e2e",
     /* Maximum time one test can run for. */
-    timeout: process.env.CI ? 60 * 1000 : 30 * 1000,
+    timeout: process.env.CI ? 10 * 1000 : 5 * 1000,
```
M-Kusumgar: i think these numbers were way too large, when i did a push with a couple tests that didnt work, it took ages because they also retry on CI

src/components/Choropleth.vue
```diff
@@ -16,80 +16,79 @@
...
             <LControl position="topright">
                 <ExcelDownloadButton />
             </LControl>
+            <LControl v-if="isLargeScreen" position="bottomleft" style="margin-left: 2rem; margin-bottom: 2rem">
+                <map-settings-menu :is-large-screen="true"></map-settings-menu>
+            </LControl>
         </LMap>
+        <div v-if="!isLargeScreen">
+            <map-settings-menu :is-large-screen="false"></map-settings-menu>
+        </div>
```
david-mears-2, line 30: On mobile, the gap looks weird because of the map's attribution:
image

david-mears-2, line 30: On mobile, the gap looks weird because of the map's attribution poking its nose through:

image

src/components/mapSettingsMenu/MenuActivator.vue
```diff
@@ -0,0 +1,90 @@
...
+    <v-theme-provider theme="dark">
+        <v-btn
+            v-if="appConfig && mapSettings.indicator"
+            role="button"
+            :class="props.isLargeScreen ? 'indicator-menu-activator-desktop' : 'indicator-menu-activator-mobile'"
+        >
+            <p class="text-wrap">
+                {{ buttonSummary }}
+            </p>
+            <v-icon :icon="menuOpen ? 'mdi-chevron-down' : 'mdi-chevron-up'" end></v-icon>
```
david-mears-2, line 22: I know you probably didn't introduce this, but to me these down and up seem back to front to what I'd expect.

david-mears-2: I'd expect `“up” state when expanded and the “down” state when collapsed` as suggested [here](https://primer.style/ui-patterns/progressive-disclosure#chevron-icon)

src/components/mapSettingsMenu/MenuActivator.vue
```diff
@@ -0,0 +1,90 @@
...
+        required: true
+    }
+});
+
+const buttonSummary = computed(() => {
+    const isGlobal = !admin0GeojsonFeature.value?.properties.COUNTRY;
+    const country = admin0GeojsonFeature.value?.properties.COUNTRY || "Global";
+    const indicator = appConfig.value.indicators[mapSettings.value.indicator].humanReadableName;
+    const adminLevel = mapSettings.value.adminLevel === 1 ? "Admin 1" : "Admin 2";
+    return isGlobal ? `${country} | ${indicator}` : `${country} | ${indicator} | ${adminLevel}`;
```
david-mears-2: This character is not vertically centered in the font we use.

Some alternative characters to the pipe `|` for your consideration:

1. `｜` - in the font used, this results in a grey line that is better vertically centered than the pipe
2. `│` - a taller, well-vertically centered one, that doesn't come out grey

I sourced these alternatives from https://en.wikipedia.org/wiki/Vertical_bar#Unicode_code_points

tests/e2e/index.spec.ts
```diff
@@ -20,13 +20,6 @@ test.describe("Index page", () => {
         await expect(await page.locator(".v-app-bar-title:has-text('DengueMap')")).toBeVisible();
     });
 
-    test("loading spinner is shown on page load", async ({ page }) => {
-        // page load is too fast sometimes producing a flaky test as
-        // loading spinner is not found, have to reload the page here
-        page.goto("/");
-        await expectLoadingSpinnerIsShownThenRemoved(page);
-    });
```
david-mears-2, line 28: Is this removed for being flaky?

tests/e2e/index.spec.ts
```diff
@@ -125,15 +118,21 @@ test.describe("Index page", () => {
         await expect(page).toHaveURL("/dengue/may24/serop9");
     });
 
-    test("indicator menu renders as expected", async ({ page }) => {
-        await page.click(".indicator-menu-activator");
-        await expect(await page.locator(".v-menu .v-list-item").count()).toBe(4);
-        const foiMenuItem = await page.locator(":nth-match(.v-menu .v-list-item, 1)");
+    test("map settings menu renders as expected global", async ({ page }) => {
```
david-mears-2:
```suggestion
    test("map settings menu renders as expected when no country is selected", async ({ page }) => {
```

tests/e2e/index.spec.ts
```diff
@@ -146,8 +145,19 @@ test.describe("Index page", () => {
         await expect(await ageGroupButtons.last().innerText()).toBe("95-99");
     });
 
+    test("map settings menu renders as expected", async ({ page }) => {
```
david-mears-2:
```suggestion
    test("map settings menu renders as expected when a country has been selected", async ({ page }) => {
```

tests/e2e/index.spec.ts
```diff
@@ -179,25 +189,6 @@ test.describe("Index page", () => {
...
-
-        await expect(await allRegions).toHaveCount(2060);
-
-        await expect(adminToggle).toHaveCount(1);
-        await adminToggle.getByRole("button", { name: "Admin 1" }).click();
-        await page.waitForURL(/dengue\/may24\/FOI\/AGO\/admin1/i);
-
-        await expect(await allRegions).toHaveCount(1915);
-    });
-
```
david-mears-2, line 200: Goodbye test. Why is it being deleted btw?

tests/e2e/routing.spec.ts
```diff
@@ -9,10 +9,13 @@ const expectIndexPage = async (
...
     featureCount: number,
-    selectedCountryFeatureCount: number
+    selectedCountryFeatureCount: number,
+    selectedCountryName: string
 ) => {
     await page.waitForURL(new RegExp(`${BASE_URL}${url}`));
-    await expect(await page.textContent(".indicator-menu-activator")).toBe(selectedIndicatorName);
+    await expect(await page.textContent(".indicator-menu-activator-desktop")).toBe(
+        `${selectedCountryName} | ${selectedIndicatorName}${selectedCountry ? " | Admin 2" : ""}`
+    );
```
david-mears-2: Maybe worth converting to a test that is agnostic as to which pipe character is used:

```suggestion
    await expect(await page.textContent(".indicator-menu-activator-desktop")).toContainText(selectedCountryName)
    await expect(await page.textContent(".indicator-menu-activator-desktop")).toContainText(selectedIndicatorName)
    if (selectedCountry) {
      await expect(await page.textContent(".indicator-menu-activator-desktop")).toContainText("Admin 2")
    } else {
      await expect(await page.textContent(".indicator-menu-activator-desktop")).not.toContainText("Admin")
    }
    );
```

tests/e2e/routing.spec.ts
```diff
@@ -54,18 +57,31 @@ test.describe("Router", () => {
...
+            "Tanzania"
+        );
     });
 
     test("browse to indicator and country at admin1 loads expected data", async ({ page }) => {
         await page.goto(`${BASE_URL}/FOI/VEN/admin1`);
-        await expect(await page.textContent(".indicator-menu-activator")).toBe("Force of infection");
+        await expect(await page.textContent(".indicator-menu-activator-desktop")).toBe(
+            "Venezuela | Force of infection | Admin 1"
+        );
```
david-mears-2: Could become a pipe-character-agnostic test

tests/e2e/routing.spec.ts
```diff
@@ -76,7 +92,9 @@ test.describe("Router", () => {
     test("is case-insensitive", async ({ page }) => {
         await page.goto("/DENGUE/May24/SEROP9/tza");
         await page.waitForURL(/\/DENGUE\/May24\/SEROP9\/tza/);
-        await expect(await page.textContent(".indicator-menu-activator")).toBe("Seroprevalence at age 9 years");
+        await expect(await page.textContent(".indicator-menu-activator-desktop")).toBe(
+            "Tanzania | Seroprevalence at age 9 years | Admin 2"
+        );
```
david-mears-2: Could become a pipe-character-agnostic test

david-mears-2 (commented):
Generally great especially the opacity method of highlighting selections. Got some quibbles on style, and questions.

src/components/mapSettingsMenu/MenuActivator.vue
david-mears-2: I prefer number 1

playwright.config.ts
EmmaLRussell: I think perhaps these values went in during bad old days before you speeded up the country selection.

src/components/mapSettingsMenu/MenuActivator.vue
EmmaLRussell: That's a drop-down menu though, rather than an expand-up. My intuition was that the chevron should point in the direction that the items will expand/contract into. Google maps does something similar: 
[Screencast from 06-06-24 10:35:47.webm](https://github.com/mrc-ide/arbomap/assets/44669576/bb2c611c-fdef-4ff4-a962-e3fadb13ad22)

But maybe 'down for expand' is also expected whichever the direction?

src/components/mapSettingsMenu/MenuActivator.vue
EmmaLRussell: I suppose another approach would be have a slightly different (but still dark) background for any text, with a little bit of padding between the 'sections'. That might look a little busy though, and probably (even) weird(er) with wrapping text.

EmmaLRussell: But also, just toning down the colour of the pipe to grey will also make it seem less distracting I think.

src/components/Choropleth.vue
EmmaLRussell: Agreed, think it will look better if we lose the gap on small screens.

src/components/mapSettingsMenu/MenuActivator.vue
david-mears-2: I don't feel strongly about it, and it's fundamentally about intuitions isn't it

src/types/resourceTypes.ts
```diff
@@ -63,7 +64,7 @@ export interface IndicatorValue {
 // Indicator values for a singe feature area - a dictionary of indicator ids to  values
 export type FeatureIndicatorValues = Dict<IndicatorValue>;
 
-export type MapFeature = Feature<Polygon | MultiPolygon>;
```
EmmaLRussell, line 66: What's the reason for this change?

tests/unit/components/adminLevelToggle.spec.ts
```diff
@@ -42,11 +46,15 @@ describe("AdminLevelToggle", () => {
...
+        vi.useFakeTimers();
+        const spyRouterPush = vi.spyOn(router, "push");
+        const { findAllByRole } = renderComponent(2, true);
         const buttons = await findAllByRole("button");
         const [admin1Button] = buttons;
         admin1Button.click();
-        expect(emitted()).toHaveProperty("change-admin-level", [[1]]);
+        vi.runAllTimers();
+        expect(spyRouterPush).toHaveBeenCalledWith(`/${APP_BASE_ROUTE}/FOI/TZA/admin1`);
+        vi.useRealTimers();
```
EmmaLRussell: Could put the timers stuff into beforeAll/afterAll

EmmaLRussell (commented):
Looks great! Worth all the bashing around of options I think.

src/components/Choropleth.vue
M-Kusumgar: done, have also removed the border on mobile as i thought it looked odd and also the border radius is set to 0 on mobile of the menu as well!

src/components/mapSettingsMenu/MenuActivator.vue
M-Kusumgar: i dont feel too strongly about this either, so will leave this as is for now, i think its more important that the arrows toggle than which particular states

src/components/mapSettingsMenu/MenuActivator.vue
M-Kusumgar: have taken this to the slack channel!

tests/e2e/index.spec.ts
M-Kusumgar: yes, it was too flaky and we shouldnt really see it much after the map settings refactor anyway as the choropleth isnt even mounted until we have all the data so i just removed this test, it was probably only passing because ci was being slow

tests/e2e/routing.spec.ts
M-Kusumgar: done

tests/e2e/routing.spec.ts
M-Kusumgar: done

tests/e2e/index.spec.ts
M-Kusumgar: apologies, i was going to add this back in but forgot, it is added back in now!

tests/unit/components/adminLevelToggle.spec.ts
M-Kusumgar: done

src/types/resourceTypes.ts
M-Kusumgar: so the feature type accepts generics like the type of geometry `Polygon | MultiPolygon` and the properties we expect them to have. Typescript requires you to specify all or no generics, which means it assumed the properties were of type `never` and that threw errors in my ide since we are accessing properties. So I just specified the most general type of property here, we could do our specific one with `ID_0`, `COUNTRY`, etc. but i feel like thatll be more effort than its worth to keep up if the scientists keep changing the names

EmmaLRussell (approved):
lgtm

david-mears-2 (approved):

PR #42, opened 2024-06-05
mrc-5445 Make e2e tests run in parallel 

M-Kusumgar:
e2e tests felt slow and considering that we are relying more on them in this project, it would be nice to get on top of this now


david-mears-2 (approved):
Thank you for doing this!

Reference:
https://playwright.dev/docs/test-sharding

package.json
```diff
@@ -14,7 +14,13 @@
     "lint": "eslint . --ignore-path .gitignore",
     "test:unit": "vitest",
     "test:coverage": "vitest --coverage",
-    "test:e2e": "playwright test"
+    "test:e2e": "playwright test",
+    "test:e2e:1": "playwright test --shard=1/6",
```
EmmaLRussell: Do you actually need these? You can pass command line args through to the npm scripts, so couldn't you just call
`npm run test:e2e -- --shard=${{ inputs.shard }}/6` from the action?

EmmaLRussell (commented):
Great stuff! Just wondering if it would be possible to tersify it a little?

Could this be tidied up a little by having 6 jobs in a single workflow rather than 6 separate workflows? Jobs are executed on separate runners and in parallel by default, aren't they, so wouldn't that get the parallelism we want? That would mean we'd just need one yml file and wouldn't have the repetition of `on`..?

package.json
M-Kusumgar: thanks, that is much better! also done it in one workflow file!

EmmaLRussell (approved):
Nice!

PR #44, opened 2024-06-06
mrc-5264: Country Select

david-mears-2:
Ideally, I would have written the tests as unit tests rather than e2e, but I was struggling to make the unit test open the list of countries to select one, and eventually gave up trying.

As the country select is being added for the sake of accessibility, the e2e test uses keyboard controls for most of the interactions.

- [x] make the e2e test pass on CI, which it does locally


david-mears-2:
I think the reason it was so hard to debug the CI test failures is because a race condition was making it incorrectly pass locally, whereas I had been assuming the problem was on CI (which was correctly failing)

david-mears-2:
I will squash these commits tomorrow

src/components/mapSettingsMenu/CountrySelector.vue
```diff
@@ -0,0 +1,69 @@
+<template>
+    <v-autocomplete
+        label="Country"
+        :model-value="mapSettings.country"
+        @update:model-value="handleChangeCountry"
+        :items="countryItems"
+        :custom-filter="customFilter"
+        placeholder="Start typing to filter"
```
EmmaLRussell: Not sure the word 'filter' is going to be too meaningful here as user isn't initially going to be looking at an unfiltered list. How about "to find country" or "to select country"?

src/components/mapSettingsMenu/CountrySelector.vue
```diff
@@ -0,0 +1,69 @@
...
+const router = useRouter();
+
+const { mapSettings, countryNames, mapLoading } = storeToRefs(useAppStore());
+
+const countryItems = computed(() => {
+    const items = Object.entries(countryNames.value).map(([code, name]) => ({
+        title: name,
+        value: code
+    }));
+    items.unshift({ title: "Global", value: "" });
```
EmmaLRussell: I wonder if we might want to give the user a more obvious way of getting to the global view from the selector than having to remember that it's called 'Global'. Could always be present in the drop down regardless of filter for example..?

tests/unit/components/mapSettingsMenu/countrySelector.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+    test("renders as expected on the global view", async () => {
+        const { findByText } = renderComponent("");
+        expect(await findByText("Global")).toBeVisible();
+    });
+
+    test("renders as expected on country view", async () => {
+        const { findByText } = renderComponent("MWI");
+        expect(await findByText("Malawi")).toBeVisible();
+    });
+});
```
EmmaLRussell, line 55: Is this the sort of thing you were trying to do here?
```
test("items appear when click input", async () => {
        await renderComponent();
        const user = userEvent.setup();
        const input = await screen.findByPlaceholderText("Start typing to filter");
        await user.click(input);
        const globalListItem = await screen.findByText("Global");
        expect(globalListItem).toBeVisible();
        expect((globalListItem as HTMLElement).className).toContain("v-list-item-title");
        const mwiListItem = await screen.findByText("Malawi");
        expect(mwiListItem).toBeVisible();
        expect((mwiListItem as HTMLElement).className).toContain("v-list-item-title");
    });
```

EmmaLRussell (commented):
Looks great!

I feel like when you first click on the selector it should either clear the text or select it all so that you can just start typing your search string then (rather than it getting appended to the name of the current country). I'm not sure people's first instinct is going to be to click the x button. 

Maybe some refinement could be done to the ordering to prefer items where the matched word is the first word  - currently e.g  typing "Re" puts "Central African Republic" before "Republic of the Congo".

As you spotted, accents are a bit of a problem - typing "Me" won't find "México". There's probably few enough countries that we could brute force work out which accented letters we need to deal with and map those to their unaccented equivalents when doing the matching. 

But all these are quibbles, which could be done in separate tickets if at all!

tests/unit/components/mapSettingsMenu/countrySelector.spec.ts
david-mears-2: Yes, but with either typing into the input or clicking the dropdown to filter the list, neither of which I could make work in the unit test.

src/components/mapSettingsMenu/CountrySelector.vue
david-mears-2: Now that the input is cleared when focused/clicked, the dropdown will immediately open, and so they will be looking at an unfiltered list. But 'find' is probably nicer.

src/components/mapSettingsMenu/CountrySelector.vue
david-mears-2: Updating so that 'Global' is always a visible option in the dropdown (which requires that it always be first).

tests/unit/components/mapSettingsMenu/countrySelector.spec.ts
david-mears-2: I'll include this test that doesn't do filtering, to avoid the perfect being the enemy of the good.

EmmaLRussell: How about this? 
```
const input = await screen.findByPlaceholderText("Start typing to find a country");
await user.type(input, "Tan");
await waitFor(() => expect(screen.queryByText("Malawi")).toBe(null));
```

david-mears-2: The issue is in locating the dropdown, I can't seem to find it / see it (it probably is failing to exist for some reason)

EmmaLRussell: I'd sort of assumed that if you could see the Malawi list item:
```
const mwiListItem = await screen.findByText("Malawi");
expect(mwiListItem).toBeVisible();
```
then that must mean that the drop down is open and showing its items.. 
Are you wanting to check for the existence of an element with a particular class name, and was that failing?

src/components/mapSettingsMenu/CountrySelector.vue
```diff
@@ -0,0 +1,123 @@
...
+};
+
+const countryItems = computed(() => {
+    const items = Object.entries(countryNames.value)
+        .sort(([, firstName], [, secondName]) => {
+            // Sort by whether the query is a prefix of the country name, so that a search for
+            // 'Republic' lists 'Republic of the Congo' before 'Central African Republic'
+            const query = normalizeCountryName(searchQuery.value);
+            const firstStartsWithQuery = normalizeCountryName(firstName).startsWith(query);
+            const secondStartsWithQuery = secondName.startsWith(query);
```
M-Kusumgar: why do we not normalise `secondName`?

M-Kusumgar (commented):
Nice, works great! got a small question in the comments. Couple of tiny remarks here:
* the country selector is a different height to the admin toggle, its a very tiny thing but if easy to fix then would be nice to have them the same height!
* if i selected a country, open map settings menu, click on country selector, then change my mind and click on indicator, the country name will not show up again until you close the menu and reopen it. I think it would be good to change country name back to the current country on blur

src/components/mapSettingsMenu/CountrySelector.vue
david-mears-2: Oversight :open_mouth:

tests/unit/components/mapSettingsMenu/countrySelector.spec.ts
david-mears-2: I was literally seeing no text that I expected from the dropdown, when using screen.debug(). But now we have something that works.

david-mears-2:
I have addressed all comments, @EmmaLRussell and @M-Kusumgar

tests/unit/components/mapSettingsMenu/countrySelector.spec.ts
david-mears-2: (Thanks, by the way!)

M-Kusumgar (approved):
Looks great!

EmmaLRussell (approved):
Looks great!

Sorry, just found a little bug that I didn't notice last time - if you select the country from the drop-down which is already selected, you get an infinite spinner. Go ahead and merge this and make another ticket for this fix if you prefer.

david-mears-2:
> Sorry, just found a little bug that I didn't notice last time - if you select the country from the drop-down which is already selected, you get an infinite spinner. Go ahead and merge this and make another ticket for this fix if you prefer.

Good spot. I've fixed this in the last commit.

PR #46, opened 2024-06-10
mrc-5438 Pre-generate global Excel files

EmmaLRussell:
Clare asked for an option for user to download a global Excel file which includes indicator 2 values as well as indicator 1. That felt like a good moment to make the global excel files pre-canned rather than generating them on the fly like the country-specific files. 

So this branch includes a script, `generate_global_excel.ts` to generate those files, which can be run from npm, It uses the same code as the app does to actually generate the files from the indicators and geojson read from the json files, to ensure consistent excel format. This required some refactoring of the existing excel code - I've split out `buildExcel.ts`, and now the new script (which runs offline) and the `useExcelDowload` composable (which runs in the app) both use this to do the actual workbook generation for global and country-specific files respetively. This required a little bit of refactoring of `useIndicatorColors` composable used by the builder code so that it no longer relies on the store directly, since there isn't one when running from the generate script.

The `useExcelDownload` composable still handles the request to download a global file, but now it simply downloads the pre-canned file (using download static file util lifted diret from packit!)

I've also included generating and flagging to the user the size of the indicator 2 file. (Maybe that's overkill though, or maybe we should do that for Brazil too!)


scripts/npm/generate_global_excel.ts
```diff
@@ -0,0 +1,78 @@
...
+writeGlobalWorkbook(level1Filename, false);
+const level2Filename = getFullFilename(excelFilename(null, AdminLevel.TWO));
+writeGlobalWorkbook(level2Filename, true);
+
+// 3. Save file sizes to report to user
+
+const getFileSizeMB = (filename: string) => {
+    const stats = fs.statSync(filename);
+    const fileSizeInBytes = stats.size;
+    const fileSizeMB = fileSizeInBytes / (1000 * 1000);
```
EmmaLRussell, line 67: Base 10 MB more standard than Base 2 these days..

src/excel/buildExcel.ts
```diff
@@ -0,0 +1,136 @@
+import { WorkBook } from "xlsx";
+import * as XLSX from "xlsx";
+import { Dict } from "../types/utilTypes";
+import { ColorType, FeatureIndicators, MapFeature } from "../types/resourceTypes";
+import { useIndicatorColors } from "../composables/useIndicatorColors";
+
+export class BuildExcel {
+    private readonly _appConfig: AppConfig;
```
M-Kusumgar, line 8: app config type isnt imported

src/excel/buildExcel.ts
```diff
@@ -0,0 +1,136 @@
...
+        admin1Geojson: Dict<MapFeature[]>,
+        admin2Geojson: Dict<MapFeature[]>
+    ) {
+        this._appConfig = appConfig;
+        this._countryNames = countryNames;
+        this._admin1Indicators = admin1Indicators;
+        this._admin2Indicators = admin2Indicators;
+        this._admin1Geojson = admin1Geojson;
+        this._admin2Geojson = admin2Geojson;
+        this._indicatorColors = useIndicatorColors(ref(appConfig));
```
M-Kusumgar: or something like that so it matches the function signature
```suggestion
        this._indicatorColors = useIndicatorColors(ref(appConfig), {} as any);
```
maybe if you dont want to use any
```suggestion
        this._indicatorColors = useIndicatorColors(ref(appConfig), computed(() => ({})));
```

src/excel/buildExcel.ts
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    private _writeTab(
+        workbook: WorkBook,
+        level: number,
+        indicatorValues: Dict<FeatureIndicators>,
+        geojson: Dict<MapFeature[]>,
+        country?: string[]
```
M-Kusumgar:
```suggestion
        country?: string
```

src/composables/useExcelDownload.ts
```diff
@@ -1,138 +1,69 @@
...
             }
         })();
     };
 
-    return { download, downloadError };
+    const downloadSelectedCountry = () => {
+        download(doDownloadSelectedCountry);
+    };
+
+    const downloadGlobal = (includeAdmin2: bool) => {
```
M-Kusumgar:
```suggestion
    const downloadGlobal = (includeAdmin2: boolean) => {
```

M-Kusumgar (approved):
Nice, looks good, like the approach and there are a lootttt of admin 2 regions wow, just got a few typescript comments but all very small!

src/excel/buildExcel.ts
M-Kusumgar: i have realised we do this in other places too, not too concerned about it haha but yh completely optional

src/composables/useExcelDownload.ts
```diff
@@ -1,138 +1,69 @@
...
+        download(async () => {
+            await doDownloadGlobal(includeAdmin2);
+        });
```
david-mears-2: Having all these similarly named functions (download, downloadGlobal, doDownloadGlobal, doDownload) gets confusing to read. I would suggest, instead of having a separate function called doDownloadGlobal, we put the implementation in downloadGlobal:

```suggestion
        download(async () => {
            const filename = excelFilename(null, includeAdmin2 ? AdminLevel.TWO : AdminLevel.ONE);
            const url = `${APP_BASE_URL}/resources/excel/${filename}`;
            await downloadFile(url, filename);
        });
```

And the equivalent for non-global (selected country).

What do you think?

README.md
```diff
@@ -46,11 +46,17 @@ the global view. Generated from the xlsx file in `data/raw` by the script `proce
...
 All public data is committed to github, and is not currently automatically updated using the scripts. So you need to 
 remember to run scripts manually when required:
 
 - if indicator values have changed: run `process-data.R`
 - if admin1 geojson has changed: run `transform_adm_geojson_to_dict.cjs`
 - if there are any changes to the list of countries, or country names: run `generate_country_names_from_adm0_geojson.cjs`
 
-Remember to copy resulting data from the `data/processed` folder to the relevant part of `public/denuge/may24/resources`.
\ No newline at end of file
+Remember to copy resulting data from the `data/processed` folder to the relevant part of `public/denuge/may24/resources`.
```
david-mears-2:
```suggestion
Remember to copy resulting data from the `data/processed` folder to the relevant part of `public/dengue/may24/resources`.
```

scripts/npm/generate_global_excel.ts
```diff
@@ -0,0 +1,78 @@
...
+    return fileSizeMB.toFixed(1);
+};
+
+const level1Size = getFileSizeMB(level1Filename);
+const level2Size = getFileSizeMB(level2Filename);
+
+const fileSizesSrc = `// NB This file is generated by generate_global_excel.ts
+export default {level1: ${level1Size}, level2: ${level2Size}}`;
```
david-mears-2:
```suggestion
export default { level1: ${level1Size}, level2: ${level2Size} }`;
```

To prevent linting complaints in the generated file.

tests/unit/components/excelDownloadButton.spec.ts
```diff
@@ -20,24 +45,53 @@ describe("ExcelDownloadButton", () => {
         expect(mockDownloadError.value).toBe(null);
     });
 
-    it("clicking button calls download", async () => {
+    it("clicking button calls download when a country is selected", async () => {
+        renderComponent(selectedCountryStore);
+        await clickButton();
```
david-mears-2, line 50:
```suggestion
        await clickButton();
        expectDialogToBeHidden();
```

david-mears-2: (maybe)

david-mears-2 (approved):

src/composables/useExcelDownload.ts
EmmaLRussell: Yeah, will do. It does mean an extra level of nesting for the eyes to negotiate and brain to perceive, but I think you're right that it's probably better than a forest of download.. methods..

PR #47, opened 2024-06-13
Update github action to Node 20 version

david-mears-2:
See https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/


M-Kusumgar (approved):
Nice

PR #49, opened 2024-11-07
Proof of concept: load map data from tile server

EmmaLRussell:
Loosely associated with ticket mrc-5382, but unlikely to be merged in its current form.. 

This branch is a proof of concept of loading map data in the form of protobuf (data) tiles rather than directly from geojson files. This means that we don't have to load all the map data in a single hit, and can load lower resolution data for wider areas at lower zoom levels. So we can serve higher resolution data at higher zoom levels than are feasible with geojson. 

This branch is currently deployed to the dev pages site: https://mrc-ide.github.io/arbomap

A separate repository will implement the tile server. The tiles it serves are generated into .mbtiles files (which are really sqlite databases).  The first client for this will probably be EasyMap (generic load and visualise spatial data) but following the pattern demonstrated here. A poc tile server has been implemented by modifying node-mbtiles as described below.

In the front end, we use VectorTileLayers (an extension of Leaflet): https://gitlab.com/jkuebart/Leaflet.VectorTileLayer
When we configure these layers, we specify the url scheme of the tile server from which to fetch tiles, and how to style the geojson returned e.g. fill colour based on indicator value for each feature. 

The Excel Download is disabled here as it currently relies on metadata in the geojson files being available, and these have been removed in this branch. We will need the tile server to provide geojson metadata before we can get that working. So as well as generating tiles from source geojson, we will need a metadata extraction script (also to sqlite db maybe) and for the tile server to provide an API to this metadata. 

The non-arbomap code bits:
1. **Prepare geojson**: Wes has provided 100% geojson. In order for tippecanoe (which generates the tiles) to give us separate layers per geojson feature (so we can style them individually), we also need to inject the `tippecanoe.layer` property with the feature id. This is done by running `scripts/tiling_poc/add_tippecanoe_layer_names.cjs`, providing input and output folders. You'd typically do this once per admin level e.g. 
`node add_tippecanoe_layer_names.cjs ~/gadm41/admin2 ~/gadm41_processed/admin2 GID_2`

2. **Generation of mbtiles**: This is done with [tippecanoe](https://github.com/mapbox/tippecanoe). Follow the installation instructions, then run `scripts/tiling_poc/generate_mbtiles` to generate the tile db for a layer from a folder of processed geojson e.g. `./scripts/tiling_poc/generate_mbtiles tiles/admin2.mbtiles ~/gadm41_processed/admin2`.
The mbtiles files are too big to include here, but are uploaded to the arbomap server, in the `tile_layers` folder.

3. **Running the server**:  I've used a modified version of [node-mbtiles](https://github.com/asd20752/node-mbtiles):
 clone this repo, npm install, then paste in the modified `main.js` from `node-mbtiles/poc` before running. This modification allows us to serve pbf tiles rather than image tiles (sets content type and encoding) and also allows CORS. Specifiy the mbtiles sources and the dataset (url path) each corresponds to when we run the server: 
```
node main.js --databases "admin0;../arbomap/tiles/admin0.mbtiles,admin1;../arbomap/tiles/admin1.mbtiles,admin2;../arbomap/tiles/admin2.mbtiles"
```
There are also docker scripts here in this branch for building, pushing and pulling a docker image with the above described server. This server is curently running on http://wpia-bots.dide.ic.ac.uk:5000 and is being proxied through https://mrcdata.dide.ic.ac.uk/grout. There's a new prop in arbomap's config.json for this tile server base url. 

**NB**
- Zoom level could be increased. Current max zoom is 10. 
- There are some 404s coming from the tile server, though it doesn't look like we have any missing data on the map. I think this must be because we don't have a full set of countries in arbomap, so there are some world tiles for which we don't have any tile data, so 404s are being returned. We'll need to generate a full set of countries for the real tile server.


src/components/Choropleth.vue
```diff
@@ -42,74 +42,91 @@ import { useIndicatorColors } from "../composables/useIndicatorColors";
...
 const { getFillAndOutlineColor } = useIndicatorColors(appConfig, selectedIndicators);
 
-const featureInSelectedCountry = (feature: MapFeature) =>
-    feature.properties[featureProperties.country] === mapSettings.value.country;
+const featureInSelectedCountry = (properties: GeoJsonProperties) =>
+    properties[featureProperties.country] === mapSettings.value.country;
 
-const featureAdminLevel = (feature: MapFeature) =>
-    featureInSelectedCountry(feature) && mapSettings.value.adminLevel === 2 ? 2 : 1;
+const featureAdminLevel = (properties: GeoJsonProperties) => {
```
EmmaLRussell, line 67: Updated this method to better match its name!

